PR #973, opened 2026-05-21
feat(server): add ParameterMatch structured filter to search API

### LINKED ISSUE #971: feat(server): add ParameterMatch structured filter to search API
## Summary

Add a `ParameterMatch` filter type to `POST /api/v1/runs/search` that allows matching specific parameter values in hook outputs without requiring raw JSONPath expressions.

## Motivation

When integrating Capsula with computation tools (e.g., veriq), the client needs to find runs where input parameters match specific values — for example, "find the latest thermal simulation run where `solar_flux ≈ 1361.0`."

The existing `HookFilter` (JSONPath-based) can achieve this, but requires the client to:
1. Know PostgreSQL JSONPath syntax
2. Manually construct range expressions for approximate matching
3. Have no server-side validation beyond JSONPath parsing

`ParameterMatch` provides a structured, validated alternative.

### Why separate from HookFilter?

| | HookFilter | ParameterMatch |
|---|---|---|
| Expressiveness | Full JSONPath syntax | Single parameter value comparison |
| Client burden | Must know JSONPath syntax | Send structured data only |
| Phase filter | None (hook_id only) | Explicit (pre/post) |
| Validation | Syntax-only | Type-level (operator + value type) |

- Backward compatible — does not change existing `HookFilter` behavior
- Room for future index optimization (e.g., BTREE on extracted JSONB values)

## Design

Each `ParameterMatch` is converted server-side to a JSONPath expression and added as an EXISTS subquery:

```
ParameterMatch { parameter: "solar_flux", operator: "ge", value: 1347.39 }
  → jsonb_path_exists(ro.output, '$.solar_flux ? (@ >= 1347.39)'::jsonpath)
```

Leverages the existing GIN index on `run_outputs.output`.

### API Types

```rust
pub enum ComparisonOp { Eq, Ne, Gt, Ge, Lt, Le }

pub struct ParameterMatch {
    pub hook_id: String,
    pub phase: String,          // "pre" or "post"
    pub parameter: String,      // dot-separated path in hook output
    pub operator: ComparisonOp,
    pub value: serde_json::Value,
}
```

### Example Request

```json
{
  "vault": "thermal-sim",
  "parameter_matches": [
    {"hook_id": "capture-command", "phase": "pre",
     "parameter": "solar_flux", "operator": "ge", "value": 1347.39},
    {"hook_id": "capture-command", "phase": "pre",
     "parameter": "solar_flux", "operator": "le", "value": 1374.61}
  ],
  "include": ["hooks"],
  "order": "latest_first",
  "limit": 1
}
```

shunsuke-shimomura:
## Summary

Adds `ParameterMatch` to `POST /api/v1/runs/search` for structured filtering over parameter-capturing hooks (`capture-json`, `capture-toml`, ...). Each entry selects `run_outputs` rows structurally (by the presence of a top-level `content` field, decoupled from `hook_id`), then narrows by any of:

- `file` — exact match on `run_outputs.config->>'path'`
- `hook_index` — 0-based position in `capsula.toml`'s phase array (disambiguates repeated `hook_id` / file combinations)
- `conditions` — one or more `<parameter> <operator> <value>` predicates, compiled into a single JSONPath and applied per row

`conditions` is a list so a range like `lr >= 0.005 && lr <= 0.05` compiles to one predicate on the same row. Two separate `parameter_matches` entries would AND across rows and can falsely match a run whose rows separately satisfy each bound.

## Example — range on the same field

```json
POST /api/v1/runs/search
{
  "vault": "training",
  "parameter_matches": [{
    "phase": "pre",
    "file": "train.json",
    "conditions": [
      { "parameter": "lr", "operator": "ge", "value": 0.005 },
      { "parameter": "lr", "operator": "le", "value": 0.05 }
    ]
  }]
}
```

compiles to `$ ? (@.content.lr >= 0.005 && @.content.lr <= 0.05)` inside a single `EXISTS` subquery.

## Changes

**`capsula-api-types`** — public wire types
- `ComparisonOp` (`eq` / `ne` / `gt` / `ge` / `lt` / `le`)
- `Phase` (`pre` / `post`), replaces raw `String`
- `ParameterCondition { parameter, operator, value }`
- `ParameterMatch { phase, file?, hook_index?, conditions[] }` — `deny_unknown_fields`
- `SearchRunsRequest.parameter_matches: Vec<ParameterMatch>`

**`capsula-server`**
- `models.rs`: `Deserialize`-only mirrors with `deny_unknown_fields`
- `query.rs`:
  - `with_parameter_match` — assembles the `EXISTS` clause and per-condition JSONPath
  - `build_conditions_jsonpath` — single condition emits the compact `$.content.<path> ? (@ <op> <val>)`; two or more emit the composable `$ ? (@.content.<p1> <op1> <v1> && ...)`
  - `build_path` — splits the parameter on `.`, quotes any segment outside bare ASCII `[A-Za-z0-9_]`, escapes `"` and `\` inside quoted segments — so keys like `learning-rate` and `温度` are addressable without opening injection
  - `QueryError::InvalidParameterMatch(String)` for structured-match errors
  - Validation: `phase` at serde level; `hook_index >= 0`; `Bool` value only with `eq`/`ne`; `MAX_PARAMETER_MATCHES = 32`; per-condition `parameter` length + empty-segment + control-char rejection; `value` restricted to number / string / boolean

## Validation semantics

| Constraint | Enforced by | Behaviour |
|---|---|---|
| Unknown `phase` string | serde | 4xx via handler |
| Outer `parameter` / `operator` typo | `deny_unknown_fields` | 4xx via handler |
| Missing at least one of `file` / `hook_index` / `conditions` | `with_parameter_match` | `InvalidParameterMatch` |
| Negative `hook_index` | `with_parameter_match` | `InvalidParameterMatch` |
| Bool value + ordering operator | `with_parameter_match` | `InvalidParameterMatch` |
| `parameter_matches.len() > 32` | `from_request` | `InvalidParameterMatch` |
| Empty / oversized / control-char in `parameter` | `build_path` | `InvalidParameterMatch` |
| Non-number/string/bool `value` | `value_to_jsonpath` | `InvalidParameterMatch` |

## Tests

Unit tests in `crates/capsula-server/src/query.rs` (25 for `pm_*`):
- shape / composition: `pm_file_and_single_condition_generate_full_query`, `pm_file_only_omits_jsonpath`, `pm_condition_only_omits_file_clause`, `pm_hook_index_only_generates_hook_index_clause`, `pm_hook_index_with_file_and_conditions_composes_all_clauses`, `pm_hook_index_alone_satisfies_at_least_one_rule`
- multi-condition: `pm_multi_condition_range_on_same_field_compiles_to_single_predicate`, `pm_multi_condition_different_fields_compose`
- path quoting: `pm_accepts_hyphenated_parameter_via_quoting`, `pm_accepts_non_ascii_parameter_via_quoting`, `pm_nested_path_mixes_bare_and_quoted_segments`, `pm_quoted_segment_escapes_backslash_and_quote`
- path rejection: `pm_rejects_empty_parameter`, `pm_rejects_empty_parameter_segment`, `pm_rejects_control_char_in_parameter`, `pm_rejects_oversize_parameter`
- value validation: `pm_string_value_is_quoted_and_escaped`, `pm_rejects_null_value`, `pm_rejects_array_value`, `pm_rejects_bool_value_with_ordering_operator`, `pm_accepts_bool_value_with_eq_and_ne`
- shape rejection: `pm_rejects_empty_match`, `pm_rejects_invalid_phase`, `pm_rejects_negative_hook_index`, `pm_rejects_too_many_parameter_matches`

Integration tests in `crates/capsula-server/tests/api_tests.rs` (22 `pm_*` tests) exercise the endpoint end-to-end against a testcontainer Postgres. Highlights:
- `pm_range_on_same_field_uses_per_row_semantics` — adversarial: one run with two rows on the same file (`lr = 0.001`, `lr = 0.1`), range `[0.005, 0.05]` must not match under the new per-row semantics
- `pm_range_query_on_same_file` — normal range on distinct-row fixtures
- `pm_outer_parameter_operator_value_rejected_as_unknown_fields` — locks in the `deny_unknown_fields` guard against the old flat-triple wire form
- `pm_condition_missing_value_is_rejected` — locks in serde-level rejection of an incomplete `ParameterCondition`

## Follow-ups (not in this PR)

- `jsonb_path_exists()` is not GIN-indexable; switching this call site and the pre-existing `with_hook_filter` call to the `@?` operator would restore index usage. Tracked separately.
- `hook_filters` has no per-request count cap; adding a parallel `MAX_HOOK_FILTERS` was scoped out.
- `SearchRunsRequest` is duplicated between `capsula-api-types` and `capsula-server::models` (pre-existing structural point).

Closes #971

### DIFF
--- a/crates/capsula-api-types/src/lib.rs
+++ b/crates/capsula-api-types/src/lib.rs
@@ -60,6 +60,83 @@ pub struct HookFilter {
     pub output_filter: String,
 }
 
+/// Comparison operator for parameter matching
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum ComparisonOp {
+    /// Exact equality
+    Eq,
+    /// Not equal
+    Ne,
+    /// Greater than
+    Gt,
+    /// Greater than or equal
+    Ge,
+    /// Less than
+    Lt,
+    /// Less than or equal
+    Le,
+}
+
+/// A structured filter for parameter-capturing hooks (e.g., `capture-json`,
+/// `capture-toml`, ...).
+///
+/// Targets `run_outputs` rows whose `output` JSON has the shape:
+///
+/// ```text
+/// { "file": "<configured-path>", "parameters": { ...parsed... } }
+/// ```
+///
+/// The server selects matching rows structurally (by the presence of both
+/// the `file` and `parameters` top-level fields) — it is decoupled from
+/// the concrete `hook_id`, so any future hook that emits this shape works
+/// automatically.
+///
+/// Constraints (validated server-side):
+///
+/// - At least one of `file` / `parameter` must be specified.
+/// - If `parameter` is present, `operator` and `value` must also be present.
+/// - If `parameter` is absent, `operator` and `value` must also be absent.
+/// - Specifying `parameter` without `file` emits a server-side warning,
+///   since the match will scan across every parameter-capturing row of
+///   the run regardless of which file it came from.
+///
+/// # Example
+///
+/// ```json
+/// {
+///     "phase": "pre",
+///     "file": "config/sat1/orbit.json",
+///     "parameter": "a",
+///     "operator": "ge",
+///     "value": 1.0
+/// }
+/// ```
+///
+/// generates `$.parameters.a ? (@ >= 1.0)` against the row whose
+/// `file == "config/sat1/orbit.json"`.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ParameterMatch {
+    /// Phase: `"pre"` or `"post"`
+    pub phase: String,
+    /// Optional exact match on the captured `file` field. Omitting this
+    /// causes the match to apply across every parameter-capturing row of
+    /// the run; the server logs a warning when this happens.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub file: Option<String>,
+    /// Dot path within the captured `parameters` object (e.g., `"lr"`,
+    /// `"sat1.orbit.a"`). When present, `operator` and `value` are required.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub parameter: Option<String>,
+    /// Comparison operator (required iff `parameter` is present).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub operator: Option<ComparisonOp>,
+    /// Value to compare against — number, string, or boolean. Required
+    /// iff `parameter` is present.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub value: Option<serde_json::Value>,
+}
+
 /// Fields that can be included in search response
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "snake_case")]
@@ -101,6 +178,11 @@ pub struct SearchRunsRequest {
     /// Hook output filters (AND logic)
     #[serde(default)]
     pub hook_filters: Vec<HookFilter>,
+    /// Structured parameter match filters for parameter-capturing hooks
+    /// (`capture-json`, `capture-toml`, ...). Each entry is an independent
+    /// EXISTS subquery; multiple entries are AND-combined.
+    #[serde(default)]
+    pub parameter_matches: Vec<ParameterMatch>,
     /// What to include in response
     #[serde(default)]
     pub include: Vec<IncludeField>,
--- a/crates/capsula-api-types/src/lib.rs
+++ b/crates/capsula-api-types/src/lib.rs
@@ -81,25 +81,28 @@ pub enum ComparisonOp {
 /// A structured filter for parameter-capturing hooks (e.g., `capture-json`,
 /// `capture-toml`, ...).
 ///
-/// Targets `run_outputs` rows whose `output` JSON has the shape:
+/// Targets `run_outputs` rows produced by hooks whose `output` has a
+/// top-level `content` field:
 ///
 /// ```text
-/// { "file": "<configured-path>", "parameters": { ...parsed... } }
+/// output: { "content": { ...parsed file contents... } }
+/// config: { "path": "<configured-path>", ... }
 /// ```
 ///
-/// The server selects matching rows structurally (by the presence of both
-/// the `file` and `parameters` top-level fields) — it is decoupled from
-/// the concrete `hook_id`, so any future hook that emits this shape works
+/// The server selects matching rows structurally (by the presence of
+/// the `content` top-level field) — decoupled from the concrete
+/// `hook_id`, so any future hook that emits this shape works
 /// automatically.
 ///
 /// Constraints (validated server-side):
 ///
-/// - At least one of `file` / `parameter` must be specified.
+/// - At least one of `file` / `hook_index` / `parameter` must be specified.
 /// - If `parameter` is present, `operator` and `value` must also be present.
 /// - If `parameter` is absent, `operator` and `value` must also be absent.
-/// - Specifying `parameter` without `file` emits a server-side warning,
-///   since the match will scan across every parameter-capturing row of
-///   the run regardless of which file it came from.
+/// - Specifying `parameter` without `file` (and without `hook_index`)
+///   emits a server-side warning, since the match will scan across
+///   every parameter-capturing row of the run regardless of which
+///   file it came from.
 ///
 /// # Example
 ///
@@ -113,18 +116,40 @@ pub enum ComparisonOp {
 /// }
 /// ```
 ///
-/// generates `$.parameters.a ? (@ >= 1.0)` against the row whose
-/// `file == "config/sat1/orbit.json"`.
+/// generates `$.content.a ? (@ >= 1.0)` against the row whose
+/// `config.path == "config/sat1/orbit.json"`.
+///
+/// # Example — using `hook_index`
+///
+/// Pin the match to the hook at position 1 (0-indexed) in the phase's
+/// array, useful when several entries share the same file / `hook_id`:
+///
+/// ```json
+/// {
+///     "phase": "pre",
+///     "hook_index": 1,
+///     "parameter": "a",
+///     "operator": "ge",
+///     "value": 1.0
+/// }
+/// ```
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct ParameterMatch {
     /// Phase: `"pre"` or `"post"`
     pub phase: String,
-    /// Optional exact match on the captured `file` field. Omitting this
-    /// causes the match to apply across every parameter-capturing row of
-    /// the run; the server logs a warning when this happens.
+    /// Optional exact match on the captured file path (compared against
+    /// `run_outputs.config->>'path'`). Omitting this causes the match to
+    /// apply across every parameter-capturing row of the run; the server
+    /// logs a warning when neither `file` nor `hook_index` is supplied.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub file: Option<String>,
-    /// Dot path within the captured `parameters` object (e.g., `"lr"`,
+    /// Optional exact match on the 0-based position of the hook in
+    /// `capsula.toml`'s `pre_run` / `post_run` list. Useful when several
+    /// entries share the same file / `hook_id` and disambiguation by
+    /// position is needed (e.g., "the 2nd `capture-json`").
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub hook_index: Option<i32>,
+    /// Dot path within the captured `content` object (e.g., `"lr"`,
     /// `"sat1.orbit.a"`). When present, `operator` and `value` are required.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub parameter: Option<String>,
--- a/crates/capsula-api-types/src/lib.rs
+++ b/crates/capsula-api-types/src/lib.rs
@@ -78,88 +78,70 @@ pub enum ComparisonOp {
     Le,
 }
 
-/// A structured filter for parameter-capturing hooks (e.g., `capture-json`,
+/// Hook execution phase.
+///
+/// Serialized as `"pre"` / `"post"` on the wire.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Phase {
+    Pre,
+    Post,
+}
+
+/// A single `<parameter> <operator> <value>` comparison inside the
+/// captured `content` object. See [`ParameterMatch::conditions`].
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ParameterCondition {
+    pub parameter: String,
+    pub operator: ComparisonOp,
+    pub value: serde_json::Value,
+}
+
+/// A structured filter over parameter-capturing hooks (`capture-json`,
 /// `capture-toml`, ...).
 ///
-/// Targets `run_outputs` rows produced by hooks whose `output` has a
-/// top-level `content` field:
+/// Selects `run_outputs` rows structurally by the presence of a
+/// top-level `content` field, then optionally pins by `file` /
+/// `hook_index` and/or filters by `conditions`.
 ///
-/// ```text
-/// output: { "content": { ...parsed file contents... } }
-/// config: { "path": "<configured-path>", ... }
-/// ```
+/// # Why `conditions` is a list, not a single `parameter` + `operator` + `value`
 ///
-/// The server selects matching rows structurally (by the presence of
-/// the `content` top-level field) — decoupled from the concrete
-/// `hook_id`, so any future hook that emits this shape works
-/// automatically.
+/// Two entries at the `parameter_matches` level are AND-combined across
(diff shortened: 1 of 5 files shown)


shunsuke-shimomura:
depends on #1008

shunichironomura:
> [!NOTE]
> This review comment was generated by AI (Claude Code). The branch was reviewed and verified locally at head 72894bc: all query-builder unit tests and the 15 new `pm_` integration tests pass (Postgres 18 via testcontainers), and clippy is clean on both touched crates.

Solid, well-tested implementation — the query construction is injection-safe by construction (all user values bound as `$N`, parameter names whitelisted *and* the assembled JSONPath re-parsed with `sql-json-path`), it mirrors the existing `with_hook_filter` conventions, and the integration tests (phase isolation, `hook_index` decoys, structural exclusion of non-parameter hooks) are thorough. A few issues before merge, plus some follow-up candidates.

## Issues

### 1. PR description is out of date

The body says each `ParameterMatch` specifies a **hook ID**, the example uses `"hook_id": "capture-command"`, and the listed test names (`test_parameter_match_numeric`, …) don't exist. The shipped design has no `hook_id` — it matches structurally (`output ? 'content'`) plus `file` / `hook_index`, as documented in [the `ParameterMatch` docs](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-api-types/src/lib.rs#L78-L137). Since the description becomes the record of the change, it should describe the final design.

### 2. Cross-row AND-of-EXISTS semantics can produce false positives (worth documenting)

Each entry compiles to an independent `EXISTS` ([query.rs#L301-L305](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L301-L305), AND-joined in [`build_query`](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L339-L349)), so a range query like `lr >= 0.005` + `lr <= 0.05` means "*some* row satisfies the lower bound AND *some (possibly different)* row satisfies the upper bound." Since the `hook_index` widening explicitly allows two `capture-json` rows with the same `config.path` in one run, a run with rows `lr=0.001` and `lr=0.1` would wrongly match that range. `pm_range_query_on_same_file` passes only because each fixture run has one row per file.

Suggest at minimum adding this caveat to [the `parameter_matches` docs](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-api-types/src/lib.rs#L206-L210). A clean future fix is multiple conditions per entry compiled into a single predicate, e.g. `$.content.lr ? (@ >= 0.005 && @ <= 0.05)`, in [`build_parameter_jsonpath`](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L398-L440).

### 3. `jsonb_path_exists()` does not use the GIN index

The description claims the filter leverages the GIN index on `run_outputs.output`, but only the operator forms `@?` / `@@` are GIN-indexable — a `jsonb_path_exists(...)` *function call* is not ([query.rs#L293-L295](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L293-L295)). The structural clause `ro.output ? 'content'` ([query.rs#L276](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L276)) *can* use it. If index leverage is the goal, write `ro.output @? $N::jsonpath` (documented as equivalent to `jsonb_path_exists` with `silent=true`). The pre-existing `with_hook_filter` has the same pattern ([query.rs#L176-L193](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L176-L193)), so this could be one follow-up covering both.

### 4. Parameter-name whitelist excludes common keys

The `[alnum, _, .]` check ([query.rs#L409-L415](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L409-L415)) means TOML/JSON keys with hyphens (`learning-rate`), spaces, etc. can never be queried. An alternative that removes the whole class: split the path on `.` and quote each segment (`$.content."learning-rate"`), escaping like the value strings, at the interpolation site ([query.rs#L439](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L439)).

Also, `char::is_alphanumeric` is Unicode-aware — broader than the `[A-Za-z0-9_.]` claimed in the doc comment ([query.rs#L403](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L403)) — so Unicode keys pass the filter but only work if they happen to parse as JSONPath identifiers. Either fix the doc or tighten to ASCII, and state the restriction in [the `parameter` field docs](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-api-types/src/lib.rs#L152-L155).

## Minor

- All validation failures reuse `QueryError::InvalidJsonPath` ([query.rs#L22-L26](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L22-L26)), so a bad phase surfaces as `"Invalid JSONPath expression: phase must be 'pre' or 'post'"` ([query.rs#L232](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L232), [L243](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L243), [L253](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L253)). An `InvalidParameterMatch(String)` variant would read much better.
- Validation errors return HTTP 200 with `{"status":"error"}` ([lib.rs#L592-L601](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/lib.rs#L592-L601)) — pre-existing handler convention, not this PR's fault; a 400 would be a worthwhile follow-up.
- Negative `hook_index` is accepted and silently matches nothing ([query.rs#L284-L288](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L284-L288)); rejecting `< 0` gives clearer feedback.
- The `tracing::warn!` for unconstrained matches ([query.rs#L260-L265](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L260-L265)) is only visible to server operators, while the guidance is aimed at the API caller who never sees it.
- The `"` / `\` escaping in string values ([query.rs#L428-L430](https://github.com/ut-issl/capsula/blob/72894bc5702691e63f48d5acb313d7dc9a6e895b/crates/capsula-server/src/query.rs#L428-L430)) has no test — `pm_build_jsonpath_string_value_quoted` only covers a plain string.

shunichironomura:
I'm still reviewing this. Meanwhile, above is the review of this PR by AI. (Doesn't mean I agree with all of the findings by AI)

shunsuke-shimomura:
Follow-up on the review above (commits `acfcc4f`…`a1147de`). AI findings that were addressed vs deferred:

## AI review — Issues

| # | Issue | Status | How |
|---|---|---|---|
| 1 | PR description out of date | **Fixed** | Rewrote the PR body to reflect the shipped design (no `hook_id`, real test names, new `conditions` shape). |
| 2 | Cross-row `AND`-of-`EXISTS` false positives on ranges | **Fixed** | `ParameterMatch` no longer has a flat `parameter`/`operator`/`value` triple. Instead it carries `conditions: Vec<ParameterCondition>`, all compiled into a single `JSONPath` predicate. Single condition → compact form `$.content.<path> ? (@ op val)`; two or more → composable `$ ? (@.content.<p1> op1 v1 && ...)`. Adversarial integration test `pm_range_on_same_field_uses_per_row_semantics` seeds one run with two rows (`lr=0.001`, `lr=0.1`) and asserts a range `[0.005, 0.05]` does *not* match — this would have false-positived under the old two-entry shape. Unit tests `pm_multi_condition_range_on_same_field_compiles_to_single_predicate` and `pm_multi_condition_different_fields_compose` lock the emitted JSONPath. |
| 3 | `jsonb_path_exists()` is not GIN-indexable; use `@?` | **Deferred** | Pre-existing `with_hook_filter` has the same pattern; per the review's own suggestion, addressed in a follow-up that covers both call sites. Not part of this PR. |
| 4 | Parameter name whitelist rejects hyphens/spaces/non-ASCII; `char::is_alphanumeric` mismatches the doc's `[A-Za-z0-9_.]` claim | **Fixed** | Replaced the flat whitelist with a segment-quoting `build_path`: parameter is split on `.`, each segment that fits bare ASCII `[A-Za-z0-9_]+` is emitted unquoted, otherwise the segment is wrapped in `"…"` with `"` / `\` escaped. Empty segments (leading/trailing/consecutive `.`), control characters, and oversize (>200 bytes) inputs are rejected explicitly. Tests `pm_accepts_hyphenated_parameter_via_quoting` (`learning-rate`), `pm_accepts_non_ascii_parameter_via_quoting` (`温度`), `pm_nested_path_mixes_bare_and_quoted_segments`, `pm_rejects_empty_parameter`, `pm_rejects_empty_parameter_segment`, `pm_rejects_control_char_in_parameter`, `pm_rejects_oversize_parameter`, `pm_quoted_segment_escapes_backslash_and_quote`. |

## AI review — Minor

| # | Issue | Status | How |
|---|---|---|---|
| M1 | All validation failures reuse `QueryError::InvalidJsonPath` | **Fixed** | Added `QueryError::InvalidParameterMatch(String)`. All `ParameterMatch` structural / semantic errors now surface with the correct variant. |
| M2 | Validation errors return HTTP 200 with `{"status":"error"}` | **Partial** | Serde-level failures (unknown `phase`, unknown outer fields, missing `ParameterCondition.value`) now return HTTP 4xx directly through axum's extractor. Handler-level validation errors (empty match, negative `hook_index`, bool + ordering op, etc.) still use the pre-existing `200 + {"status":"error"}` envelope — that's a cross-cutting handler convention beyond this PR's scope. |
| M3 | Negative `hook_index` silently matches nothing | **Fixed** | `with_parameter_match` rejects `hook_index < 0` with `InvalidParameterMatch`. Test `pm_rejects_negative_hook_index`. |
| M4 | `tracing::warn!` only visible to operators; the guidance targets the API caller | **Deferred** | Surfacing warnings on the wire is a broader response-shape change; not part of this PR. |
| M5 | `pm_build_jsonpath_string_value_quoted` covers only a plain string — no `"` / `\` escape test | **Fixed** | Added `pm_string_value_is_quoted_and_escaped` (value side) and `pm_quoted_segment_escapes_backslash_and_quote` (parameter side); both feed the emitted expression back through `sql-json-path::JsonPath::new` to prove it round-trips. |

## Fixes beyond the AI review

Prompted by a second local pass while addressing the items above.

- `phase: String` → `pub enum Phase { Pre, Post }` with `#[serde(rename_all = "snake_case")]`. Unknown values are rejected at deserialize time instead of surfacing as `"Invalid JSONPath expression: phase must be 'pre' or 'post'"` from the runtime match. Test `pm_rejects_invalid_phase` now hits the serde layer directly.
- `#[serde(deny_unknown_fields)]` on `ParameterMatch` and `ParameterCondition`. Without it, a client still sending the old flat triple would silently be interpreted as a file-only match. Test `pm_outer_parameter_operator_value_rejected_as_unknown_fields`.
- `MAX_PARAMETER_MATCHES = 32` — each entry compiles to an independent `EXISTS`, and the unbounded list was a DoS surface. Enforced in `from_request`; test `pm_rejects_too_many_parameter_matches`.
- Bool `value` with ordering operator (`Gt`/`Ge`/`Lt`/`Le`) — rejected up-front instead of erroring out inside Postgres at query time. Tests `pm_rejects_bool_value_with_ordering_operator` (4 operators) and the positive `pm_accepts_bool_value_with_eq_and_ne`.
- Non-scalar `value` (null / array / object) — explicit rejection through `value_to_jsonpath`. Tests `pm_rejects_null_value`, `pm_rejects_array_value`.
- `MAX_JSONPATH_LENGTH` doc corrected — the limit is on `str::len()` (bytes), not code points as previous prose implied.

## Known limitations acknowledged in the PR body

- `hook_filters` has no per-request count cap (pre-existing).
- `SearchRunsRequest` is duplicated between `capsula-api-types` and `capsula-server::models` (pre-existing structural point).
- Handler-level validation errors return HTTP 200 with a `status:"error"` body (see M2).

Verification: 25 `pm_*` unit tests pass locally; CI green on `a1147de` (clippy, fmt, doc, check, SQLx, MSRV, coverage, typos, zizmor).

shunsuke-shimomura:
Closing without merging.

On reflection, `ParameterMatch` is essentially syntactic sugar over the existing `hook_filters` / `output_filter`: the same query can already be expressed as a hand-written JSONPath through that field. The extra API surface, validation logic, and test/doc burden aren't justified by the ergonomic gain. If we later want a nicer surface for parameter-capturing hooks specifically, we can revisit it with the hindsight of concrete usage of `hook_filters`.

PR #1008, opened 2026-06-07
feat(capture-parameter): add hook for queryable structured parameter capture

shunsuke-shimomura:
## Summary

Adds a new hook `capture-parameter` that parses TOML / JSON / YAML configuration
files, normalizes them to JSON, and embeds the parsed values directly into
`pre-run.json` / `post-run.json` so that the capsula-server can query them with
sql-json-path. Whereas `capture-file` preserves bytes for archival / audit,
`capture-parameter` preserves structured values for cross-run query / analysis.

## Motivation

`capture-file` records the byte-exact state of a config file, but the server
can only index its `hash` and `path`. There is no way to answer queries like
"find every run with `learning_rate > 0.01`" across runs. With
`capture-parameter`:

- Hyperparameter sweeps become queryable from the server with SQL JSON path.
- Configs written in different formats (TOML / JSON / YAML) are normalized
  into a single queryable schema.
- Dashboards can join run-level parameters with metrics.

### Configuration

```toml
[[pre-run.hooks]]
id = "capture-parameter"
glob = "config/**/*.{json,yaml}"   # file selection
strip_prefix = "config"            # optional namespace prefix to strip
```

### Directory hierarchy → nested keys

`config/sat1/orbit.json` (containing `{"a": 1}`), with
`strip_prefix = "config"`, becomes queryable as `parameters.sat1.orbit.a`.
The intermediate directory components plus the file stem form the nested
key path.

### Merge & conflict semantics

When multiple files land at the same key path, their contents are
**deep-merged**. `ParameterConflict` is raised only when values genuinely
disagree. This unifies several situations that would otherwise need separate
handling:

| Situation | Outcome |
|---|---|
| `orbit.json` + `orbit.yaml` (disjoint keys) | merged |
| `orbit.json` + `orbit.yaml` (same key, same value) | OK (idempotent) |
| `orbit.json` + `orbit.yaml` (same key, different value) | `ParameterConflict` |
| `sat1.json` (leaf) and `sat1/orbit.json` (intermediate node) | merged |
| Differing arrays at the same key | `ParameterConflict` (no element-wise merge) |

### Output example

Input layout:

```
config/sat1/orbit.json     { "a": 1, "b": 2 }
config/sat1/attitude.yaml  { "q": [0, 0, 0, 1] }
config/sat2/orbit.json     { "a": 3 }
```

Captured output (excerpt of `pre-run.json`):

```json
{
  "__meta": { "id": "capture-parameter", "config": {...}, "success": true },
  "parameters": {
    "sat1": {
      "attitude": { "q": [0, 0, 0, 1] },
      "orbit":    { "a": 1, "b": 2 }
    },
    "sat2": {
      "orbit": { "a": 3 }
    }
  }
}
```

### New crate

- `crates/capsula-capture-parameter/` — `ParameterHook`, `ParameterCaptured`,
  helper functions (`compute_keys`, `merge_value`, `merge_at_keys`),
  `ParameterHookError`.

### Existing crates

- `crates/capsula-registry/` — registers `ParameterHook` in both the pre-run
  and post-run registries.
- `crates/capsula-capture-file/src/lib.rs` — `mod error;` → `pub mod error;`
  so that `FileHookError` can be wrapped from `capture-parameter` via
  `#[error(transparent)]`.
- `Cargo.toml` (workspace) — adds `capsula-capture-parameter`; bumps the
  workspace version from `0.12.1` to **`0.13.0`** (minor bump for the new
  feature; backward compatible).

### Documentation

- `docs/hooks/capture-parameter.md` (new):
  - Configuration options (`glob`, `strip_prefix`)
  - Output example showing directory hierarchy → nested keys
  - **Merge & Conflict Semantics** section
  - When to use `capture-parameter` vs `capture-file` (two-axis comparison
    table, scenario decision matrix, combined-usage pattern)
  - Limitations covering JSON-incompatible information (TOML datetime, YAML
    anchors, etc.)

## Tests

`cargo test -p capsula-capture-parameter` — **23 tests, all green**.

### Unit tests (15) — `src/lib.rs` `#[cfg(test)] mod tests`
- `compute_keys`: simple / nested / strip_prefix / strip_prefix mismatch
- `merge_value`: disjoint objects, equal scalars, conflicting scalars, type
  mismatch, equal arrays, differing arrays, deep object merge
- `merge_at_keys`: leaf insertion, nested object creation, leaf↔intermediate
  unification, conflict path reporting

### Integration tests (8) — `tests/integration.rs`
- Single JSON / TOML file parsing
- Same-stem cross-format merge (`orbit.json` + `orbit.yaml`)
- Nested directories with `strip_prefix`
- Leaf and intermediate node merge (`sat1.json` + `sat1/orbit.json`)
- `ParameterConflict` on value disagreement
- Unsupported extensions (e.g., `.csv`) error
- `strip_prefix` mismatch error

## Limitations

Because the value is normalized to JSON, the following are lost. Use
`capture-file` alongside if these matter:

- Comments in TOML / YAML
- TOML datetime type tags (coerced to RFC 3339 strings)
- YAML anchors / aliases (resolved at parse time; references become duplicated)
- YAML non-string mapping keys (fails JSON serialization)
- YAML `.inf` / `.nan` (not representable in JSON)
- Element-wise array merge (arrays are treated as opaque leaves)

## Test Plan

- [x] `cargo test -p capsula-capture-parameter` — 23 tests pass
- [x] `cargo build --release -p capsula`
- [x] End-to-end demo in `/home/shimomura/capsula-demo` — a Monte Carlo Pi
      estimator with `capture-file` and `capture-parameter` applied to the
      same glob, output diff confirmed
- [x] CI green

## Breaking Changes

None. New hook only; existing capsula configurations are unaffected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -521,7 +521,7 @@ checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "capsula"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
@@ -544,7 +544,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-api-types"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "serde",
  "serde_json",
@@ -552,7 +552,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-capture-command"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "serde",
@@ -564,7 +564,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-capture-cwd"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "serde",
@@ -576,7 +576,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-capture-env"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "serde",
@@ -588,7 +588,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-capture-file"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "glob",
@@ -602,7 +602,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-capture-git-repo"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "git2",
@@ -615,7 +615,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-capture-machine"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "serde",
@@ -627,8 +627,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "capsula-capture-parameter"
+version = "0.13.0"
+dependencies = [
+ "capsula-capture-file",
+ "capsula-core",
+ "glob",
+ "serde",
+ "serde_json",
+ "serde_yaml",
+ "tempfile",
+ "thiserror 2.0.18",
+ "toml",
+ "tracing",
+ "ulid",
+]
+
+[[package]]
 name = "capsula-client"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-api-types",
  "reqwest",
@@ -640,7 +657,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-config"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "capsula-core",
  "capsula-registry",
@@ -653,7 +670,7 @@ dependencies = [
 
 [[package]]
 name = "capsula-core"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
  "chrono",
  "clap",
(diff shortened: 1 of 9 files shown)


shunichironomura:
Thank you for the PR.

It seems to me this design provides too much functionality (and thus brings complexity) without justifiable benefit. Can you explain why simple `capture-{json,yaml,toml}` hooks that capture a single file is not sufficient?

Specifically, judging from the PR description alone, this PR supports:

1. Capturing multiple files via glob pattern
2. Merging the contents of the captured files
3. Stripping the prefix of the accessor

Can you provide the rationale for each of them?
In my opinion Item 1 is the most justifiable of the three, while Item 3 is the least.

shunsuke-shimomura:
Thank you for the feedback.

Regarding item 3, I was also unsure whether it was really necessary, so I implemented it as an optional configuration field. The reason I added it is that I assumed it could be useful when the nesting becomes too deep, or when the directory containing the config files is moved. In such cases, changing only this option would allow users to keep using the same queries.

Regarding item 2, I also considered returning parameter objects as an array, one per file. However, there are cases where the directory structure of the config files itself carries semantic meaning for the parameters. In such cases, I thought it would often be more appropriate to merge the directory structure into a single object.

For example:

https://github.com/Akira-nkjm/craft/blob/main/systems/aocs/data.toml
https://github.com/Akira-nkjm/craft/blob/main/systems/cdh/data.toml

If these files contain components with the same name, they would conflict, since the filenames are also the same.

Regarding item 1, I thought it would simply be convenient to have, and that, together with item 2, it would make the captured data easier to query.

PR #1016, opened 2026-06-10
feat(capture-json): add hook that parses a single JSON file into queryable parameters

### LINKED ISSUE #987: `capture-json` hook

shunsuke-shimomura:
Implements #987.

## Summary

Adds a new built-in hook `capture-json` that parses a single JSON file and
embeds its parsed content in the run output under `parameters`, alongside
the configured path in `file`. By design this hook captures exactly one
file per instance — compose multiple `[[pre-run.hooks]]` (or post-run)
entries to capture several config files.

## Why one file per instance

This shape mirrors capsula's compositional design (combine many small
hooks to express the desired capture), and keeps the output schema
trivially queryable from the server: the `parameters` field is the
exact parsed JSON, so `$.parameters.lr ? (@ >= 0.01)` against a single
`run_outputs.output` row works without any wrapping layer.

## Configuration

```toml
[[pre-run.hooks]]
id = "capture-json"
path = "config/sat1/orbit.json"
```

`path` is taken relative to `project_root` (absolute paths also accepted).

## Output shape

```json
{
  "__meta": {
    "id": "capture-json",
    "config": { "path": "config/sat1/orbit.json" },
    "success": true
  },
  "file": "config/sat1/orbit.json",
  "parameters": { "a": 1.42, "b": "LEO" }
}
```

- `file` preserves the path **verbatim** as written in the config, so
  multiple `capture-json` instances with the same basename remain
  distinguishable (`config/sat1/orbit.json` vs `config/sat2/orbit.json`).
- `parameters` contains the parsed JSON, ready for sql-json-path queries.

## Changes

- new crate: `crates/capsula-capture-json/` (Hook, Config, Captured, errors)
- registered in `capsula-registry` for both `PreRun` and `PostRun` phases
- workspace `Cargo.toml`: added `capsula-capture-json`
- docs: `docs/hooks/capture-json.md`

## Tests

6 unit tests in the crate (`cargo test -p capsula-capture-json`):

- `parses_valid_json_into_parameters_field`
- `file_field_preserves_configured_path_verbatim`
- `supports_nested_json_objects`
- `missing_file_returns_io_error`
- `invalid_json_returns_parse_error`
- `serialized_output_has_file_and_parameters_fields`

## Verification

- [x] `cargo build -p capsula-capture-json -p capsula-registry`
- [x] `cargo test -p capsula-capture-json` — 6/6 pass
- [x] `cargo clippy -p capsula-capture-json -p capsula-registry --all-targets -- -D warnings`
- [x] `cargo fmt --check`

## Follow-ups (separate PRs)

This PR is one of a small set replacing the abandoned single-PR
\`capture-parameter\` approach (#1008). Companion hooks (e.g. \`capture-yaml\`,
\`capture-toml\`) will follow in their own PRs with the same output shape.

### DIFF
--- a/crates/capsula-capture-json/src/lib.rs
+++ b/crates/capsula-capture-json/src/lib.rs
@@ -1,9 +1,11 @@
 //! `capture-json` hook: parse a single JSON file and embed its parsed
-//! content in the run output under the `parameters` field.
+//! content in the run output under the `content` field.
 //!
 //! By design this hook captures exactly one file per instance. Compose
 //! multiple `capture-json` entries in `capsula.toml` to capture multiple
-//! files.
+//! files. The path written in the config is preserved as part of the
+//! standard `__meta.config.path`, so the captured value does not need
+//! to duplicate it.
 
 mod error;
 
@@ -28,17 +30,12 @@ pub struct JsonHookConfig {
 #[derive(Debug)]
 pub struct JsonHook {
     config: JsonHookConfig,
-    project_root: PathBuf,
 }
 
 #[derive(Debug, Serialize)]
 pub struct JsonCaptured {
-    /// The path as written in the configuration (verbatim, including any
-    /// directory components). Useful for distinguishing multiple
-    /// `capture-json` outputs that share the same basename.
-    file: String,
     /// The parsed content of the file as JSON.
-    parameters: serde_json::Value,
+    content: serde_json::Value,
 }
 
 impl<P> Hook<P> for JsonHook
@@ -50,13 +47,10 @@ where
     type Config = JsonHookConfig;
     type Output = JsonCaptured;
 
-    fn from_config(config: &serde_json::Value, project_root: &Path) -> CapsulaResult<Self> {
+    fn from_config(config: &serde_json::Value, _project_root: &Path) -> CapsulaResult<Self> {
         let config: JsonHookConfig =
             serde_json::from_value(config.clone()).map_err(JsonHookError::from)?;
-        Ok(Self {
-            config,
-            project_root: project_root.to_path_buf(),
-        })
+        Ok(Self { config })
     }
 
     fn config(&self) -> &Self::Config {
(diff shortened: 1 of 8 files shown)


crates/capsula-capture-json/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+#[derive(Debug, Serialize)]
+pub struct JsonCaptured {
+    /// The path as written in the configuration (verbatim, including any
+    /// directory components). Useful for distinguishing multiple
+    /// `capture-json` outputs that share the same basename.
+    file: String,
+    /// The parsed content of the file as JSON.
+    parameters: serde_json::Value,
```
shunichironomura: I think we should have a better name than `parameters`, because what we want to capture using `capture-json` hook is not always parameters.

I think `content` is a good option as it is more general-purpose.

crates/capsula-capture-json/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+    config: JsonHookConfig,
+    project_root: PathBuf,
+}
+
+#[derive(Debug, Serialize)]
+pub struct JsonCaptured {
+    /// The path as written in the configuration (verbatim, including any
+    /// directory components). Useful for distinguishing multiple
+    /// `capture-json` outputs that share the same basename.
+    file: String,
```
shunichironomura: If this is equivalent to the `path` in the configuration, we don't need this as it would be redundant. Or does it carry extra information?

crates/capsula-capture-json/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+pub struct JsonHookConfig {
+    /// Path to the JSON file to parse, relative to `project_root`.
+    /// Absolute paths are also accepted.
+    path: PathBuf,
+}
+
+#[derive(Debug)]
+pub struct JsonHook {
+    config: JsonHookConfig,
+    project_root: PathBuf,
```
shunichironomura: This is not wrong, but, strictly speaking, not necessary, as we can access the project root via `metadata.project_root` in the `Hook::run` method:

https://github.com/ut-issl/capsula/blob/10f7e4e80b194d7570b0fc07afec13342c248c2c/crates/capsula-capture-file/src/lib.rs#L110-L115

crates/capsula-capture-json/src/lib.rs
shunsuke-shimomura: My rough thinking here was that querying metadata would be a bit cumbersome, but if the duplication is more problematic, I’m happy to refactor it.

shunichironomura: I don't think querying the configuration field is cumbersome enough to duplicate, it would look something like this:
```json
{
  "hook_filters": [
    {
      "hook_id": "capture-json",
      "config_filter": "$.path ? (@ == \"config/sat1/orbit.json\")",
      "output_filter": "$.content.sat1.orbit.a ? (@ == 1.42)"
    }
  ],
}
```

shunichironomura (approved):

PR #1019, opened 2026-06-10
fix(server): widen run_outputs uniqueness to include config_hash

### LINKED ISSUE #1017: Server upsert overwrites multiple hooks with same hook_id in one phase
> [!NOTE]
> This bug is reported by Codex and not yet verified by human. 

## Summary

Capsula server stores hook outputs with a uniqueness/upsert key of `(run_id, phase, hook_id)`. This causes multiple hook instances with the same `hook_id` in the same phase to overwrite each other during upload.

This is especially visible for the new `capture-json` hook, which is designed to support multiple entries in `capsula.toml` to capture multiple JSON files. Only one `capture-json` output survives on the server.

## Impact

- Runs can contain multiple hook outputs locally in `pre-run.json` / `post-run.json`.
- After `capsula push`, the server retains only one output per `(run_id, phase, hook_id)`.
- Querying by `config_filter` such as `$.path ? (@ == "config/sat1/orbit.json")` cannot find overwritten hook instances.
- Capturing a `file` field in the hook output does not solve this because the entire row is overwritten.

## Expected behavior

The server should preserve every configured hook instance for a run/phase, even when multiple instances share the same `hook_id`.

## Relevant code

- `crates/capsula-server/migrations/20260411000000_add_unique_constraint_run_outputs.sql` creates `idx_run_outputs_unique` on `(run_id, phase, hook_id)`.
- `crates/capsula-server/src/lib.rs` uploads hooks with `ON CONFLICT (run_id, phase, hook_id) DO UPDATE`.

## Possible fixes

- Include an instance discriminator in the uniqueness key, e.g. config hash, hook index, or explicit hook instance id.
- Or allow duplicate `run_outputs` rows for the same `(run_id, phase, hook_id)` and make repeated uploads idempotent by another key.

shunsuke-shimomura:
Closes #1017.

## Problem

`run_outputs` had `UNIQUE(run_id, phase, hook_id)`. The upload handler
used that key as the `ON CONFLICT` target, so a second hook with the
same `hook_id` in the same phase silently **overwrote** the first.

This was visible in the new `capture-json` / `capture-toml` hooks, which
are explicitly designed to be registered multiple times in one phase —
one entry per file to capture. Only the last upload survived.

## Fix

Add a generated `config_hash` column derived from the canonical jsonb
text of `config`, and widen the uniqueness key to include it:

```sql
ALTER TABLE run_outputs
ADD COLUMN config_hash TEXT
GENERATED ALWAYS AS (md5(coalesce(config::text, ''))) STORED;

DROP INDEX idx_run_outputs_unique;
CREATE UNIQUE INDEX idx_run_outputs_unique
ON run_outputs(run_id, phase, hook_id, config_hash);
```

Postgres canonicalizes jsonb text (sorted keys, no whitespace), so the
hash is stable across re-uploads of the same config — the runtime never
has to compute the hash itself.

Both upload INSERT statements switch to
`ON CONFLICT (run_id, phase, hook_id, config_hash) DO UPDATE`, with
`config = EXCLUDED.config` dropped from the SET clause (a matching
config_hash implies an equivalent config, so it is a no-op).

Existing rows are unique under the old key, so the generated column
backfills deterministically without any duplicate-resolution step.

## Tests

New integration test `multiple_hook_instances_with_same_id_coexist`
exercises three uploads against a real Postgres via testcontainers:

1. Upload two `capture-json` hooks with identical `hook_id` but distinct
   `config.path` values → both rows persist.
2. Fetch the run → both `content` payloads come back.
3. Re-upload one of the two with the *same* config but a different
   `content` → existing row is UPSERTed (no duplicate created); the
   other row is untouched.

## Verification

- [x] `cargo test -p capsula-server --test api_tests multiple_hook_instances_with_same_id_coexist`
- [x] `cargo test -p capsula-server --test api_tests test_hook_outputs_storage` (existing regression — unchanged)
- [x] `cargo sqlx prepare --workspace --check` clean — `.sqlx/` cache is checked in

## Notes

- The migration is forward-only and safe to apply to an existing
  database: existing rows are already unique under the old key, so the
  generated column does not introduce conflicts.
- The `.sqlx/` offline query cache is regenerated and committed to keep
  the CI `sqlx prepare --check` job green.

### DIFF
--- a/.sqlx/query-0f041bcff67becde780d8f50cd520ce827c65019c919c96f3bc748818b962057.json
+++ b/.sqlx/query-0f041bcff67becde780d8f50cd520ce827c65019c919c96f3bc748818b962057.json
@@ -1,0 +1,88 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n                SELECT phase, hook_id, config, output, success, error\n                FROM run_outputs\n                WHERE run_id = $1\n                ORDER BY id\n                ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "phase",
+        "type_info": "Text",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "phase"
+          }
+        }
+      },
+      {
+        "ordinal": 1,
+        "name": "hook_id",
+        "type_info": "Text",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "hook_id"
+          }
+        }
+      },
+      {
+        "ordinal": 2,
+        "name": "config",
+        "type_info": "Jsonb",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "config"
+          }
+        }
+      },
+      {
+        "ordinal": 3,
+        "name": "output",
+        "type_info": "Jsonb",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "output"
+          }
+        }
+      },
+      {
+        "ordinal": 4,
+        "name": "success",
+        "type_info": "Bool",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "success"
+          }
+        }
+      },
+      {
+        "ordinal": 5,
+        "name": "error",
+        "type_info": "Text",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "error"
+          }
+        }
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      true,
+      false,
+      false,
+      true
+    ]
+  },
+  "hash": "0f041bcff67becde780d8f50cd520ce827c65019c919c96f3bc748818b962057"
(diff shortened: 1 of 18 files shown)


shunichironomura:
@shunsuke-shimomura Thank you for the fix. I really appreciate your effort. However, I think the correct approach here is to use the hook index (0, 1, ...) rather than the config hash, just like we do for the artifact directory naming:

```
.capsula/my-project/2025-01-09/143022-happy-river/
├── _capsula/
│   ├── metadata.json           # What ran, when, and where
│   ├── pre-run.json            # Environment before
│   ├── command.json            # Command output
│   └── post-run.json           # Results after
├── pre-0-capture-git-repo/     # Per-hook artifact directory
└── post-0-capture-file/        # Per-hook artifact directory
    └── output.txt              # Your output file
```

This is because it *is* theoretically possible to use the same configuration twice with the different semantics when, for example, you:

1. Capture file `config.json` via `capture-file` hook
2. Run some command using `capture-command` hook, modifying `config.json`
3. Capture file `config.json` via `capture-file` hook again

You would have the exact same config for the two `capture-file` hooks, but the output will have different semantics and thus should be stored separately.

PR #1036, opened 2026-06-19
fix(server): key run_outputs by hook_index, allowing duplicate hook_id rows

### LINKED ISSUE #1017: Server upsert overwrites multiple hooks with same hook_id in one phase

shunsuke-shimomura:
Refs #1017. Replaces the closed #1019 (which used a config hash and was rejected because two hooks with the same hook_id AND same config — e.g. two `capture-command` rows running the identical command — would still collide).

## Approach

Use each hook's **global position** in the capsula.toml `pre_run` / `post_run` array as the discriminator. The upload handler iterates each incoming JSON array with `.enumerate()` and supplies `hook_index` on every INSERT. The unique key tightens to `(run_id, phase, hook_index)` and no longer mentions `hook_id`.

This is what the original issue suggested ("hook index, or explicit hook instance id") and what the reviewer pointed toward. It works regardless of whether the colliding hooks have distinct or identical configs.

## Why drop `hook_id` from the unique key

The user pointed out that querying on a per-`hook_id` index is awkward, while a global `hook_index` lets you write natural SQL:

```sql
-- Show hooks in original capsula.toml order
ORDER BY phase, hook_index;

-- Find "the 3rd pre-run hook"
WHERE phase = 'pre' AND hook_index = 2;
```

`hook_id` stays as a regular column for filtering / display but no longer participates in the uniqueness constraint.

## Migration

```sql
ALTER TABLE run_outputs
ADD COLUMN hook_index INTEGER NOT NULL DEFAULT 0;

-- Backfill: assign sequential hook_index within each (run_id, phase)
-- using insertion order (id) as proxy for original array position.
UPDATE run_outputs r SET hook_index = o.new_index
FROM (
    SELECT id,
           ROW_NUMBER() OVER (PARTITION BY run_id, phase ORDER BY id) - 1 AS new_index
    FROM run_outputs
) AS o
WHERE r.id = o.id;

DROP INDEX idx_run_outputs_unique;
CREATE UNIQUE INDEX idx_run_outputs_unique
ON run_outputs(run_id, phase, hook_index);
```

Under the previous `UNIQUE(run_id, phase, hook_id)` there was at most one row per `(run_id, phase, hook_id)`, so the backfill assignment is unambiguous and re-uploads of historical runs continue to UPSERT correctly.

## Runtime

Both upload `INSERT` statements (pre and post phases) now:

- Use `for (idx, hook) in hooks.iter().enumerate()` to compute `hook_index`.
- Bind `hook_index` as `$7`.
- Use `ON CONFLICT (run_id, phase, hook_index)` (no `hook_id`).
- `SET hook_id = EXCLUDED.hook_id` is included so re-uploads with an edited `capsula.toml` (different hook at the same position) reflect the new id.

## Tests

Three integration tests against a real Postgres via testcontainers:

1. `multiple_hooks_with_distinct_configs_coexist` — the original #1017 case (two `capture-json` with different paths). Both rows persist.
2. `multiple_hooks_with_identical_configs_coexist` — the reviewer's concern that closed #1019 (two `capture-command` with the **same** command, distinguished only by position). Both rows persist with distinct `stdout`s.
3. `re_upload_with_same_array_is_idempotent` — re-uploading the same array structure UPSERTs existing rows instead of duplicating, and updates `content` to the latest payload.

## Verification

- [x] `cargo build -p capsula-server`
- [x] `cargo test -p capsula-server --test api_tests multiple_hooks` — 2/2 pass
- [x] `cargo test -p capsula-server --test api_tests re_upload` — 1/1 pass
- [x] `cargo test -p capsula-server --test api_tests test_hook_outputs_storage` — existing regression unchanged
- [x] `.sqlx/` cache regenerated and checked in

## Notes

The migration is forward-only and safe to apply to an existing database: existing rows are already unique under the old key, the backfill is deterministic, and the new index covers them without duplicate-resolution work.

### DIFF
--- a/.sqlx/query-0f041bcff67becde780d8f50cd520ce827c65019c919c96f3bc748818b962057.json
+++ b/.sqlx/query-0f041bcff67becde780d8f50cd520ce827c65019c919c96f3bc748818b962057.json
@@ -1,0 +1,88 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n                SELECT phase, hook_id, config, output, success, error\n                FROM run_outputs\n                WHERE run_id = $1\n                ORDER BY id\n                ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "phase",
+        "type_info": "Text",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "phase"
+          }
+        }
+      },
+      {
+        "ordinal": 1,
+        "name": "hook_id",
+        "type_info": "Text",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "hook_id"
+          }
+        }
+      },
+      {
+        "ordinal": 2,
+        "name": "config",
+        "type_info": "Jsonb",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "config"
+          }
+        }
+      },
+      {
+        "ordinal": 3,
+        "name": "output",
+        "type_info": "Jsonb",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "output"
+          }
+        }
+      },
+      {
+        "ordinal": 4,
+        "name": "success",
+        "type_info": "Bool",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "success"
+          }
+        }
+      },
+      {
+        "ordinal": 5,
+        "name": "error",
+        "type_info": "Text",
+        "origin": {
+          "Table": {
+            "table": "run_outputs",
+            "name": "error"
+          }
+        }
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      true,
+      false,
+      false,
+      true
+    ]
+  },
+  "hash": "0f041bcff67becde780d8f50cd520ce827c65019c919c96f3bc748818b962057"
+}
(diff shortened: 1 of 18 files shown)


shunichironomura (approved):
Thank you!

Migration _may_ add arbitrary indexes irrelevant to the original order in the config file, but it is fine as long as we use index only for conflict avoidance, which means we should have different means like tagging if we want to different hooks with the same config in the query.

PR #1097, opened 2026-07-07
Fail closed on hook failures

### LINKED ISSUE #1062: Abort gates fail open: a hook *error* never aborts the run (`allow_dirty`/`require_pushed`/`abort_on_failure` bypassed)
## Summary

All run-blocking safety gates are signaled through `Captured::abort_requested()`, which only exists when a hook **succeeds**. If a gate hook *errors*, the error is recorded as non-fatal and the wrapped command runs anyway — the gate fails open.

## Details

`crates/capsula-orchestration/src/hooks.rs:94-105` — the error branch maps every hook error to `abort = false`:

```rust
Err(e) => {
    error!("Failed to run {hook_identifier} ...: {error:#}");
    let json = json!({ "__meta": { "success": false, "error": ... }});
    (json, false)   // <- never aborts
}
```

The abort flag is only produced on the `Ok` path (`captured.abort_requested()`, line 79). Affected gates:

- `capture-git-repo` with `allow_dirty = false` / `require_pushed = true` — the abort request only exists if `GitHook::run` returns `Ok`. Any error inside it (artifact-dir creation failure, a `tag_lightweight` name collision, a libgit2/permissions error) is swallowed and the run proceeds against a dirty/unpushed repo. The tag-collision path is especially reachable: `crates/capsula-capture-git-repo/src/lib.rs:171-177` calls `tag_lightweight(..., force=false)` and run names come from a small random word-pair space.
- `capture-command` with `abort_on_failure = true` — if the guard command fails to spawn (missing binary → `ExecutionFailed`), the error is non-fatal and the run proceeds.

## Failure scenario

A CI job relies on `allow_dirty = false` to refuse running on a dirty tree. A transient error in the git hook (e.g. tag name already exists from an earlier run with the same generated name) turns the hard gate into a no-op, and the command executes on the dirty tree with `success: false` merely recorded in `pre-run.json`.

## Suggested fix

Gates that can block execution must fail closed. Options: give hooks a way to declare "errors are fatal" (an abort-capable marker), or treat an error from any hook that *can* request abort as an abort request. At minimum, a hook error on a configured gate should not be downgraded to a warning.

Related: exit-code/abort propagation in the CLI (separate issue).

shunichironomura:
> [!WARNING]
> We should not merge this PR until #1099 is merged.

## Summary

This PR changes hook execution from a "best effort / fail open" model to a "fail closed after collecting diagnostics" model.

Internally, it introduces `HookOutcome<T>` so hooks can distinguish:

- `Ok(HookOutcome::Succeeded(output))`: the hook ran and its configured success condition passed
- `Ok(HookOutcome::Failed { output, reason })`: the hook ran and captured useful output, but its configured success condition failed
- `Err(error)`: the hook could not complete because of an operational error

`Hook::run` now returns `CapsulaResult<HookOutcome<Self::Output>>`, while `Captured` is only responsible for JSON serialization. All in-tree hooks, including the newly added `capture-json` and `capture-toml` hooks, use the new outcome API. The server preserves policy-failure reasons through upload, database storage, API responses, and the run-detail UI.

## AI assistance

- AI used: yes — AI coding assistants helped implement and test the changes and update this PR description.
- Human review of AI-assisted code: pending

## Breaking changes

- [ ] No breaking changes
- [x] Breaking changes; the `breaking change` label is applied

### Pre-run hook failures now stop the wrapped command

Before this PR, hook errors were recorded in `pre-run.json` as `success: false`, but the wrapped command still ran. That meant safety gates could fail open.

After this PR:

1. Capsula still runs all configured pre-run hooks to collect as much context as possible.
2. Capsula writes `pre-run.json` with every hook result.
3. If any pre-run hook fails or errors, Capsula exits with the hook-failure exit code before running the wrapped command.

This affects both policy failures and operational errors:

- dirty git repo when `capture-git-repo.allow_dirty = false`
- unpushed git commit when `capture-git-repo.require_pushed = true`
- a pre-run hook returning an error, such as a missing guard command or failed artifact-dir creation

### Post-run hook failures now make Capsula fail after recording results

Post-run hooks cannot stop the wrapped command because it has already run. They now fail closed at the Capsula layer instead:

1. Capsula still runs all configured post-run hooks.
2. Capsula writes `post-run.json` with every hook result.
3. If any post-run hook fails or errors:
   - `capsula run` exits with the hook-failure exit code when the wrapped command succeeded.
   - `capsula run` preserves the wrapped command's non-zero exit code when the wrapped command already failed.
   - `capsula run-end` exits with the hook-failure exit code after writing `post-run.json`.

This makes post-run capture/notification failures visible to automation without masking an already-failing wrapped command.

### Hook policy failures are now visible in hook metadata

When a hook completes but its configured condition fails, the hook output is still recorded, but `__meta.success` is now `false` and `__meta.failure_reason` explains why.

For example, a dirty repo with `allow_dirty = false` is recorded as a failed hook outcome rather than a successful capture with an out-of-band abort flag.

Operational hook errors continue to be recorded with `__meta.error`.

### `capture-command` now treats non-zero exit status as failure by default

Before this PR, `capture-command` only requested abort on non-zero status when `abort_on_failure = true`.

After this PR, the default successful status set is `[0]`. A non-zero status is therefore a hook failure unless configured otherwise.

To intentionally check for a failing command, configure the expected status explicitly:

```toml
[[pre-run.hooks]]
id = "capture-command"
command = ["test", "-f", "missing-file.txt"]
success_codes = [1]
```

For compatibility, existing configs that explicitly set `abort_on_failure = false` still accept any exit status. New configs should prefer `success_codes`.

## Notes for hook developers

Hook implementations now need to return `CapsulaResult<HookOutcome<Self::Output>>` from `Hook::run`.

Use the three outcome categories intentionally:

- Return `Ok(HookOutcome::success(output))` when the hook completed and the captured state satisfies the hook's configured policy.
- Return `Ok(HookOutcome::failure(output, reason))` when the hook completed and produced useful output, but the captured state violates the hook's configured policy.
  - Example: `capture-git-repo` successfully captured repo state, but `allow_dirty = false` and the repo is dirty.
  - Example: `capture-command` successfully captured stdout/stderr/status, but the exit status is not in `success_codes`.
- Return `Err(error)` when the hook itself could not complete the operation it was asked to perform.
  - Example: command spawn failure, missing artifact directory, unreadable file, libgit2 operation failure, Slack API failure.

`Captured` no longer has `abort_requested()`. Abort/failure policy should be expressed through `HookOutcome::failure(...)`, not hidden inside captured output.

The orchestrator will still attempt every hook in the phase. Hook developers do not need to short-circuit other hooks; they only need to classify their own result correctly.

For hook configuration design, prefer explicit success criteria over ad-hoc abort booleans. For example, `capture-command` now uses `success_codes` so expected non-zero statuses can be represented directly.

## Notes for reviewers

- This is intentionally both an internal API refactor and a user-facing behavior change.
- Hooks in both phases are still all attempted before Capsula decides the phase failed.
- `capsula run` does not mask a wrapped command's non-zero exit code with a post-run hook failure code.

## Tests

- `just lint`
- `just test`

Closes #1062

---
This is **part 1 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #1147
- <kbd>&nbsp;3&nbsp;</kbd> #1146
- <kbd>&nbsp;2&nbsp;</kbd> #1098
- <kbd>&nbsp;1&nbsp;</kbd> #1097 👈


shunichironomura:
@shunsuke-shimomura FYI

shunsuke-shimomura:
Heads-up on the server side: `capsula-server`'s `HookMeta` in [`crates/capsula-server/src/models.rs`](../blob/main/crates/capsula-server/src/models.rs#L50-L56) does not have a `failure_reason` field yet, so the new `__meta.failure_reason` emitted by this PR will be silently dropped when the server ingests `pre-run.json` / `post-run.json`.

`__meta.error` (operational error, `Err` case) is already covered, but `failure_reason` (policy-violation case, `Ok(HookOutcome::Failed { reason })`) is a distinct field carrying information the server currently cannot surface — e.g. "aborted because repo was dirty" vs "aborted because git tag creation errored" would look the same in the UI/API without it.

Suggest adding:

```rust
pub struct HookMeta {
    pub id: String,
    pub config: Option<JsonValue>,
    pub success: bool,
    pub error: Option<String>,
    pub failure_reason: Option<String>, // new
}
```

and threading it through `RunOutputRow` / the DB schema / the response types so the failure reason survives the round-trip. Doesn't need to block this PR, but worth a follow-up so the extra diagnostic isn't lost on the server.

PR #974, opened 2026-05-21
feat: add Python client package for Capsula server API

shunsuke-shimomura:
## Summary

Add a typed Python client package (`clients/python/capsula-client`) for the Capsula server HTTP API.

Mirrors the Rust `capsula-client` crate, providing Python tools with first-class access to the Capsula server.

### `clients/python/capsula-client`

- `CapsulaClient`: HTTP client (`httpx`) with `list_vaults`, `vault_exists`, `search_runs`, `get_run`
- Model types mirroring `capsula-api-types`:
  - `SearchRunsRequest`, `SearchRunsResponse`, `SearchRunResult`
  - `ParameterMatch`, `ComparisonOp`, `HookFilter`, `SortOrder`
  - `VaultInfo`, `HookOutput`, `FileInfo`

## Usage

```python
from capsula_client import CapsulaClient, SearchRunsRequest, ParameterMatch, ComparisonOp

with CapsulaClient("https://capsula.example.com") as client:
    vaults = client.list_vaults()

    resp = client.search_runs(SearchRunsRequest(
        vault="thermal-sim",
        parameter_matches=[
            ParameterMatch("hook-id", "pre", "solar_flux", ComparisonOp.GE, 1347.39),
            ParameterMatch("hook-id", "pre", "solar_flux", ComparisonOp.LE, 1374.61),
        ],
        include=["hooks"],
        limit=1,
    ))
```

## Dependencies

- `httpx>=0.28`
- `pytest>=8.0` + `pytest-httpx>=0.35` (dev)

## Tests

12 tests: model serialization (6) + mocked HTTP client (6)

## Note

This PR includes the ParameterMatch server changes from #973 as a dependency. The `ParameterMatch` model type in the Python client mirrors the server-side type.

Closes #972

PR #1099, opened 2026-07-07
feat(server): expose hook_index in run responses; split HookOutput type

shunsuke-shimomura:
## Motivation

After #1036 landed the `hook_index` column on `run_outputs`, the DB knows which slot in `capsula.toml`'s `pre_run` / `post_run` array each hook came from — but the server didn't expose that to clients, and hooks came back in DB-default order. That means:

- Callers can't tell "1st `capture-command` vs 2nd `capture-command`" apart in the response.
- List order was implementation-defined, so any UI that renders hooks side-by-side had a subtle non-determinism.

## Approach

Split the previous dual-purpose `HookOutput` into two intent-named types:

| Type | Direction | serde | Has `hook_index`? |
|---|---|---|---|
| `HookOutputUploaded` / `HookMetaUploaded` | `POST /api/v1/upload` (from CLI) | `Deserialize` | No — server assigns via `.enumerate()` |
| `HookOutputQueried` / `HookMetaQueried` | GET run details, search hits | `Serialize` | Yes |

The upload type dropping `hook_index` is a small but nice invariant: the client cannot forge or omit an authoritative field, and the server side no longer needs a `#[serde(default)]` compat shim.

All three hook-fetching SELECTs now pull `hook_index` and use `ORDER BY phase, hook_index`, so the response list mirrors `capsula.toml`'s array order.

## Tests

New integration test `response_exposes_hook_index_and_preserves_array_order`:

- Uploads a 3-element `pre_run` array: `[capture-command, capture-json, capture-command]`.
- Fetches `GET /api/v1/runs/{id}`.
- Asserts `hooks[i].__meta.hook_index == [0, 1, 2]` and that content-level ordering (stdout / content payload) matches the sent array.

Existing regression tests (`test_hook_outputs_storage`, `multiple_hooks_with_distinct_configs_coexist`, `multiple_hooks_with_identical_configs_coexist`) continue to pass.

## Follow-ups

- **Python client (#974)**: needs a small mirror to include `hook_index: int` in its `HookMeta`. Will land as a separate PR.
- **Query-side hook_index filter (Layer B)**: adding `hook_index: Option<i32>` to `ParameterMatch` so "the Nth capture" can be pinned in queries. Separate PR on top of this one.

## Verification

- [x] `cargo test -p capsula-server --test api_tests response_exposes_hook_index_and_preserves_array_order` — new test passes
- [x] `cargo test -p capsula-server --test api_tests multiple_hooks` — 2/2 existing regressions pass
- [x] `cargo test -p capsula-server --test api_tests test_hook_outputs_storage` — existing hook storage test passes
- [x] `.sqlx/` offline cache regenerated and committed

### DIFF
--- a/crates/capsula-server/src/models.rs
+++ b/crates/capsula-server/src/models.rs
@@ -48,13 +48,13 @@ pub struct ListRunsQuery {
 #[derive(Debug, Deserialize)]
 pub struct HookOutputUpload {
     #[serde(rename = "__meta")]
-    pub meta: HookMetaUploaded,
+    pub meta: HookMetaUpload,
     #[serde(flatten)]
     pub output: JsonValue,
 }
 
 #[derive(Debug, Deserialize)]
-pub struct HookMetaUploaded {
+pub struct HookMetaUpload {
     pub id: String,
     pub config: Option<JsonValue>,
     pub success: bool,
@@ -70,13 +70,13 @@ pub struct HookMetaUploaded {
 #[derive(Debug, Serialize)]
 pub struct HookOutputResponse {
     #[serde(rename = "__meta")]
-    pub meta: HookMetaQueried,
+    pub meta: HookMetaResponse,
     #[serde(flatten)]
     pub output: JsonValue,
 }
 
 #[derive(Debug, Serialize)]
-pub struct HookMetaQueried {
+pub struct HookMetaResponse {
     pub id: String,
     pub config: Option<JsonValue>,
     pub success: bool,
--- a/crates/capsula-server/src/models.rs
+++ b/crates/capsula-server/src/models.rs
@@ -43,7 +43,7 @@ pub struct ListRunsQuery {
 ///
 /// The uploaded variant does not carry a `hook_index` because the server
 /// authoritatively assigns positions from each phase's array on receipt
-/// (via `.enumerate()`). See [`HookOutputQueried`] for the outbound
+/// (via `.enumerate()`). See [`HookOutputResponse`] for the outbound
 /// counterpart which does carry `hook_index`.
 #[derive(Debug, Deserialize)]
 pub struct HookOutputUploaded {
@@ -68,7 +68,7 @@ pub struct HookMetaUploaded {
 /// invocation of the same `hook_id`. See [`HookOutputUploaded`] for the
 /// inbound counterpart which omits it.
 #[derive(Debug, Serialize)]
-pub struct HookOutputQueried {
+pub struct HookOutputResponse {
     #[serde(rename = "__meta")]
     pub meta: HookMetaQueried,
     #[serde(flatten)]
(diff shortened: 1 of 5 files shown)


shunsuke-shimomura:
@shunichironomura 
I’m not too confident about the naming, so please let me know if you have any suggestions there as well.

shunsuke-shimomura:
https://github.com/ut-issl/capsula/pull/1097#issuecomment-4904199709
Also, if I apply this comment to #1097 , this PR will probably conflict with it.

shunichironomura:
@shunsuke-shimomura I don't care much about the naming that is not end-user-facing. That being said, my preference (which LLM happens to agree with) is:
- `HookOutputUpload`/`HookMetaUpload`
- `HookOutputResponse`/`HookMetaResponse`

shunichironomura:
> [#1097 (comment)](https://github.com/ut-issl/capsula/pull/1097#issuecomment-4904199709) Also, if I apply this comment to #1097 , this PR will probably conflict with it.

@shunsuke-shimomura #1097 is a relatively large internal change while this PR is a small correctness fix. So if this PR can be ready soon, it can be merged first.

shunsuke-shimomura:
change `HookOutputQueried` to `HookOutputResponse`

crates/capsula-server/src/models.rs
```diff
@@ -39,25 +39,57 @@ pub struct ListRunsQuery {
...
+#[derive(Debug, Serialize)]
+pub struct HookOutputResponse {
+    #[serde(rename = "__meta")]
+    pub meta: HookMetaQueried,
+    #[serde(flatten)]
+    pub output: JsonValue,
+}
+
+#[derive(Debug, Serialize)]
+pub struct HookMetaQueried {
```
shunichironomura:
```suggestion
pub struct HookMetaResponse {
```

crates/capsula-server/src/models.rs
```diff
@@ -39,25 +39,57 @@ pub struct ListRunsQuery {
...
-#[derive(Debug, Serialize, Deserialize)]
-pub struct HookOutput {
+/// A hook output as sent by the CLI to `POST /api/v1/upload`.
+///
+/// The uploaded variant does not carry a `hook_index` because the server
+/// authoritatively assigns positions from each phase's array on receipt
+/// (via `.enumerate()`). See [`HookOutputResponse`] for the outbound
+/// counterpart which does carry `hook_index`.
+#[derive(Debug, Deserialize)]
+pub struct HookOutputUploaded {
```
shunichironomura:
```suggestion
pub struct HookOutputUpload {
```

crates/capsula-server/src/models.rs
```diff
@@ -39,25 +39,57 @@ pub struct ListRunsQuery {
...
-    pub meta: HookMeta,
+    pub meta: HookMetaUploaded,
     #[serde(flatten)]
     pub output: JsonValue,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
-pub struct HookMeta {
+#[derive(Debug, Deserialize)]
+pub struct HookMetaUploaded {
```
shunichironomura:
```suggestion
pub struct HookMetaUpload {
```

shunichironomura (changes requested):
I renamed some struct based on my previous feedback. Note that we should change their occurrences in other locations as well, otherwise it won't compile. Apart from that it looks good to me.

shunichironomura (approved):

PR #1149, opened 2026-07-22
feat: add capsula-capture-yaml hook

### LINKED ISSUE #1148: feat: add capsula-capture-yaml hook (like capture-json / capture-toml)
## Summary

Add a new hook crate `capsula-capture-yaml`, analogous to the existing `capsula-capture-json` and `capsula-capture-toml` hooks. It should parse a single YAML file and embed its parsed content in the run output under the `content` field.

## Motivation

JSON and TOML config/metadata files can already be captured in structured form, but YAML files (common for CI configs, Kubernetes manifests, ML experiment configs, etc.) can currently only be captured as raw files via `capture-file`. A `capture-yaml` hook would make their content queryable in run outputs just like JSON/TOML.

## Design

Follow the existing pattern (see `crates/capsula-capture-toml/src/lib.rs`):

- New crate `crates/capsula-capture-yaml` with `lib.rs` + `error.rs`
- `YamlHookConfig { path: PathBuf }` with `#[serde(deny_unknown_fields)]` — one file per hook instance, relative to `project_root` (absolute paths also accepted)
- `YamlHook` implementing `Hook<P>` for both `PreRun` and `PostRun` with `ID = "capture-yaml"`
- `YamlCaptured { content: serde_json::Value }` — parsed YAML converted to JSON
- Register in `capsula-registry` via `.with_hook::<YamlHook>()` in both the pre-run and post-run registries

### Decisions

- **Parser**: use the `serde_yaml` family for a minimal implementation (deserialize straight into `serde_json::Value`). Keep it minimal for now.
- **YAML features without a JSON equivalent** (non-string mapping keys, tagged values, multi-document streams, anchors/merge keys): out of scope. Explicitly document that only fairly flat/simple YAML is supported; revisit when a real need arises.

## Tasks

- [ ] Create `capsula-capture-yaml` crate implementing `Hook<P>` and `Captured`
- [ ] Register the hook in `capsula-registry` (pre-run and post-run)
- [ ] Unit/integration tests (valid YAML, invalid YAML, missing file)
- [ ] Document the supported YAML subset (flat/simple documents only)
- [ ] Update docs (README / hook list) if applicable

shunsuke-shimomura:
> [!NOTE]
> This PR was implemented by AI (Claude Code) and has been reviewed by a human (@shimomura-shunsuke) before submission.

Closes #1148

## Summary

Adds a `capture-yaml` hook that parses a single YAML file and embeds its parsed content in the run output under the `content` field, mirroring the existing `capture-json` / `capture-toml` hooks.

- New crate `capsula-capture-yaml`, following the `capsula-capture-toml` structure (`YamlHookConfig { path }`, `YamlHook` implementing `Hook<P>` for both phases with `ID = "capture-yaml"`, `YamlCaptured { content }`)
- Uses [`yaml_serde`](https://crates.io/crates/yaml_serde) (serde_yaml maintained by The YAML Organization) — `serde_yaml` itself is unmaintained
- Registers the hook in `capsula-registry` for both pre-run and post-run phases
- Adds `docs/hooks/capture-yaml.md` and cross-links from the json/toml pages
- Docs nav (`zensical.toml`): adds `capture-yaml`, and also adds `capture-json` / `capture-toml`, which were missing from the nav

## Capture Contract

Following the review discussion (see comments below), the hook does not add its own validation layer on top of the parser. The capture contract is defined as exactly what `yaml_serde`'s untyped deserialization into `serde_json::Value` produces. The edge-case behaviors are documented in `docs/hooks/capture-yaml.md` and each is pinned by a regression test:

| YAML input | Behaviour |
| --- | --- |
| Multi-document stream (`---`) | Parse error |
| Anchors / aliases | Expanded |
| Merge keys (`<<: *a`) | **Not applied** — captured as a literal `"<<"` key |
| Non-string scalar keys (`1:`, `true:`) | Stringified; collisions are last-write-wins (silently drops one value) |
| `!!binary` values | Captured as the raw base64 string |
| Custom tags (`!Custom`) / non-scalar keys | Parse error |
| `.nan` / `±.inf` | JSON `null` |

Stricter enforcement (e.g. switching to `serde-saphyr`) remains possible later without changing the hook's config surface.

No server-side changes are needed: the ParameterMatch filter does not hardcode hook IDs, so `capture-yaml` rows are queryable as-is.

## Test Plan

- [x] 15 unit tests in `capsula-capture-yaml`, including behavior-pinning tests for every edge-case row above
- [x] `cargo test --workspace` passes on latest `main`
- [x] `cargo clippy --workspace --all-targets --all-features` / `--no-default-features` clean, `cargo fmt --check --all` clean, `cargo doc --workspace --no-deps` clean
- [x] End-to-end: `capsula run` with a `capture-yaml` entry in `capsula.toml` produces the expected `pre-run.json` row (`__meta.id = "capture-yaml"`, parsed `content`)


shunichironomura:
Below is the AI-assisted review of this PR.

Although most of the findings are about rare edge cases, if we could handle them correctly by simply using [`serde-saphyr`](https://github.com/bourumir-wyngs/serde-saphyr) instead of `yaml_serde`, I think it is a justifiable investment.

---

> [!NOTE]
> The following report was written by AI.

## Summary

The crate, registry wiring, documentation navigation, baseline error handling, and ordinary YAML tests follow the existing `capture-json` / `capture-toml` patterns well. The main issue is the direct `yaml_serde` deserialization into `serde_json::Value` at `crates/capsula-capture-yaml/src/lib.rs:80`. That conversion does not enforce the advertised “single-document, JSON-representable YAML only” boundary and can report success after changing or discarding YAML semantics.

A follow-up dependency evaluation found that [`serde-saphyr`](https://docs.rs/serde-saphyr/) can enforce most of this boundary directly: it offers duplicate-key detection, configurable merge-key handling, alias/resource limits, and strict handling of keys that YAML would infer as non-string scalars. Switching parsers can therefore address the two high-severity findings without implementing YAML semantics in Capsula. An explicit application policy is still required for custom tags because untyped deserialization intentionally discards them.

## Findings

### 1. High — Non-string mapping keys can collide and silently discard captured data

**Location:** `crates/capsula-capture-yaml/src/lib.rs:80`

`yaml_serde::from_str::<serde_json::Value>` asks Serde to deserialize YAML map keys as JSON strings. Scalar YAML keys are therefore stringified instead of rejected. Distinct YAML keys can collapse to one JSON key:

```yaml
1: numeric
"1": string
```

The hook succeeds and produces:

```json
{"1":"string"}
```

Reversing the two input entries produces `{"1":"numeric"}`. One value is silently lost in either case.

This conflicts with issue #1148, which puts non-string mapping keys out of scope, and undermines a capture/auditing tool's data-integrity guarantee. Unsupported input should fail rather than be recorded as a successful but lossy capture.

**Suggested fix:** Prefer `serde-saphyr` with `DuplicateKeyPolicy::Error` and `no_schema: true`. The former rejects duplicate and stringification-collision cases; the latter rejects plain keys such as `1` or `true` when a string is required, while quoted keys remain valid JSON object keys. Add tests for integer, boolean, sequence, duplicate, and colliding string/non-string keys.

If retaining `yaml_serde` is preferred for its official YAML-organization provenance, parse into `yaml_serde::Value` and perform a small fallible JSON-compatibility conversion as a fallback. A custom converter is not necessary if the `serde-saphyr` policy is accepted.

### 2. High — YAML merge keys succeed with incorrect semantics instead of being applied or rejected

**Location:** `crates/capsula-capture-yaml/src/lib.rs:80`

For example:

```yaml
defaults: &defaults
  retries: 3
job:
  <<: *defaults
  timeout: 5
```

The hook succeeds with this shape:

```json
{
  "defaults": {"retries": 3},
  "job": {
    "<<": {"retries": 3},
    "timeout": 5
  }
}
```

The effective YAML value of `job` contains `retries: 3`; `<<` is a merge directive, not an ordinary application key. Issue #1148 explicitly lists anchors/merge keys as out of scope, so a parse error would satisfy the agreed design. Alternatively, merge semantics could be intentionally supported. The current behavior does neither and silently records a misleading structure, which is particularly relevant to CI and experiment configuration files where anchors and merges are common.

**Suggested fix:** Choose and enforce one policy through `serde-saphyr`:

1. Use `MergeKeyPolicy::Error` to reject merge keys with a located parse error, matching issue #1148; or
2. Use `MergeKeyPolicy::Merge` to apply merge semantics in the parser, then document/test precedence behavior.

Plain aliases can be deliberately supported through parser expansion with configurable alias and parsing budgets rather than custom implementation. Add regression tests covering a single merge, merge sequences, local-key precedence, plain aliases, and the selected merge policy.

### 3. Medium — The documented rejection of `!!binary` tags is false

**Locations:**

- `docs/hooks/capture-yaml.md:101-106`
- `crates/capsula-capture-yaml/src/lib.rs:12-14`
- `crates/capsula-capture-yaml/src/lib.rs:80`

The documentation says tagged values such as `!!binary` are rejected as parse errors. In practice:

```yaml
payload: !!binary SGVsbG8=
```

succeeds and becomes:

```json
{"payload":"SGVsbG8="}
```

Custom tags such as `!Custom` do fail, but `!!binary` is accepted as an ordinary string. Users therefore cannot rely on the stated unsupported-feature boundary, and the output loses the distinction between binary-tagged data and a normal string.

**Suggested fix:** Decide whether standard binary semantics are desirable. With `serde-saphyr`, `!!binary SGVsbG8=` is base64-decoded to the UTF-8 string `"Hello"`; invalid base64 and decoded non-UTF-8 data fail when the target is `serde_json::Value`. If that policy is acceptable, document and test it rather than claiming all tags are rejected.

A parser switch alone does not enforce strict rejection of every tag: `serde-saphyr` intentionally discards custom tags such as `!Custom` during untyped `serde_json::Value` deserialization. If Capsula must reject all explicit/custom tags, retain a small tag-aware validation step. Otherwise document that custom tags do not survive the JSON conversion.

## Follow-up Library Evaluation

### Preferred option: `serde-saphyr`

The closest library-level solution is `serde-saphyr`. A configuration matching issue #1148 would be approximately:

```rust
use serde_saphyr::{DuplicateKeyPolicy, MergeKeyPolicy};

let options = serde_saphyr::options! {
    duplicate_keys: DuplicateKeyPolicy::Error,
    merge_keys: MergeKeyPolicy::Error,
    no_schema: true,
    strict_booleans: true,
};

let content = serde_saphyr::from_str_with_options::<serde_json::Value>(
    &raw,
    options,
)?;
```

Use `MergeKeyPolicy::Merge` instead if Capsula elects to support merge directives. The library also provides configurable alias replay limits and parsing budgets, which are relevant when processing untrusted or unusually large YAML.

`serde-saphyr` is pure Rust and has `#![forbid(unsafe_code)]`, fuzzing/Miri coverage, configurable resource limits, and broad YAML test-suite coverage. It is nevertheless newer than the `serde_yaml` family: the stable version evaluated was `0.0.29`, while upstream is approaching 1.0 through release candidates. It is not maintained by the YAML organization.

### Conservative alternative: retain `yaml_serde`

`yaml_serde` is the official YAML-organization-maintained fork of `serde_yaml` and has stronger institutional provenance. Its YAML-native `Value` preserves arbitrary mapping keys and most tags, and `Value::apply_merge()` delegates merge semantics to the library. Retaining it would require a small pure `yaml_serde::Value -> serde_json::Value` compatibility conversion to reject non-string keys. It does not expose the same strict policy controls, and `!!binary` tag information is already lost during parsing.

No parser can decide Capsula's complete policy automatically because YAML supports constructs JSON cannot represent, including arbitrary mapping keys and application-defined tags. The parser can own YAML syntax, aliases, merges, duplicate detection, and resource protection; Capsula must still choose and document what its JSON capture contract means.

## Revised Implementation Plan

1. **Define the capture contract:** Decide whether merges are rejected or expanded, whether aliases are supported, and whether custom tags are discarded or rejected. Specify behavior for `!!binary`, duplicate keys, non-string keys, multiple documents, and non-finite floats.
2. **Adopt the parser policy:** Prefer `serde-saphyr` and deserialize directly into `serde_json::Value` using explicit `DuplicateKeyPolicy::Error`, `MergeKeyPolicy`, `no_schema: true`, and `strict_booleans: true` options.
3. **Set resource limits:** Review and, where appropriate, explicitly configure alias replay and parser budgets for files handled by the capture hook.
4. **Handle only the remaining tag policy:** If all custom tags must be rejected, add a focused tag-aware validation pass. Do not implement merge expansion or general YAML conversion in Capsula when the parser already provides it.
5. **Expand tests:** Add regression cases for all three findings, duplicate keys, aliases, merge precedence or rejection, `!!binary`, custom tags, non-finite floats, and an automated registry/CLI-level test proving `capture-yaml` is available in an actual run.
6. **Align documentation:** Update the supported-subset table and error behavior to exactly match the selected and tested parser options.
7. **Re-run validation:** Workspace tests, both workspace Clippy configurations, formatting, and docs.

## Validation Performed

The following passed on the applied PR branch:

- `cargo test --workspace`
- `cargo test -p capsula-capture-yaml -p capsula-registry`
- `cargo clippy -p capsula-capture-yaml -p capsula-registry --all-targets --all-features`
- `cargo fmt --check --all`
- `cargo doc --workspace --no-deps`

An isolated probe using the locked `yaml_serde 0.10.4` and the same `from_str::<serde_json::Value>` call confirmed:

- Merge keys remain as literal `"<<"` keys.
- Integer and boolean map keys are stringified.
- `1` and `"1"` collide with last-write-wins data loss.
- `!!binary` is accepted as a string.
- Custom tags and sequence-valued map keys fail.
- `.nan` and `±.inf` become JSON `null` as documented.

A follow-up isolated probe using `serde-saphyr 0.0.29` confirmed:

- Merge keys are correctly expanded by default and rejected with `MergeKeyPolicy::Error`.
- Plain aliases are expanded.
- Duplicate keys and the `1` / `"1"` collision are rejected.
- `no_schema: true` rejects integer and boolean keys that would otherwise be stringified.
- Single-document deserialization rejects multiple documents.
- `!!binary SGVsbG8=` is decoded to `"Hello"`.
- Custom tags are discarded when deserializing into `serde_json::Value`, including with `no_schema: true`; strict tag rejection therefore remains an application-level decision.

## Positive Notes

- The hook is registered for both phases through the shared registry builder.
- Path resolution and non-fatal hook error propagation match existing hooks.
- Unknown configuration fields are rejected.
- The docs navigation also fixes the prior omission of `capture-json` and `capture-toml`.
- Existing unit tests cover normal nested input, missing files, invalid syntax, multi-document rejection, output shape, and project-root resolution.

shunichironomura:
Another approach is to properly document the edge-case behaviors of yaml-serde and accept its limitation.

shunsuke-shimomura:
> [!NOTE]
> This reply and the follow-up commit were drafted with AI assistance and reviewed by a human.

Thanks for the thorough review — the probe results match what we observed, and the point that the documented boundary wasn't actually enforced (in particular the false "tagged values are rejected" claim) is fair.

We went with the approach from your follow-up comment: keep `yaml_serde` and define the capture contract as exactly what its untyped deserialization into `serde_json::Value` produces, with the edge-case behaviors documented and pinned by regression tests (fc973b1):

- Merge keys are captured as literal `"<<"` keys (not applied); anchors/aliases are expanded
- Non-string scalar keys are stringified; stringification collisions are last-write-wins, now explicitly documented as silent data loss
- `!!binary` is captured as the raw base64 string; custom tags and non-scalar keys are parse errors
- Multi-document streams are parse errors

Rationale for not switching to `serde-saphyr` right now:

- Fully enforcing a "JSON-representable YAML only" boundary means either a validation layer in Capsula or a parser switch, and `serde-saphyr` is still pre-1.0 (0.0.29). I'd rather revisit once it stabilizes.
- The files this hook captures are project-authored configs rather than adversarial input, so the collision/merge cases should be pathological in practice — and they are now a documented, tested part of the contract instead of a silent surprise.
- A later switch (or a tag-aware validation pass) would not change the hook's config surface, so it can be done as a follow-up if stricter guarantees become necessary.

The crate docs and `docs/hooks/capture-yaml.md` now describe the actual behavior instead of claiming rejection, and the PR description has been updated to match.

docs/hooks/capture-yaml.md
```diff
@@ -0,0 +1,141 @@
...
+- Record the exact config used by a script in a structured, indexable form
+- Compose multiple `capture-yaml` entries to capture several config files
+
+## Configuration
+
+### Required Options
+
+| Option | Type   | Description                                                |
+| ------ | ------ | ---------------------------------------------------------- |
+| `path` | string | Path to the YAML file to parse, relative to project root.  |
```
shunichironomura: Absolute paths are also accepted

docs/hooks/capture-yaml.md
```diff
@@ -0,0 +1,141 @@
...
+| Mapping (`key: value`)        | JSON object                                    |
+
+Note that unlike TOML, YAML has no dedicated datetime type; timestamps
+written as plain scalars are captured as strings.
+
+### Edge-Case Behaviour
+
+YAML features without a JSON equivalent follow `yaml_serde`'s behavior.
+Each row below is pinned by a regression test in the
+`capsula-capture-yaml` crate:
```
shunichironomura: "Pinned by a regression test" implies that we guarantee these behaviors, but we don't want to.
```suggestion
The table below illustrates some notable behaviors of the parser provided by the `yaml_serde` crate. It is not exhaustive and these behaviors are not guaranteed by Capsula.
```

shunichironomura (changes requested):
Thank you for the PR. The implementation is good, I want some fixes to the docs.

shunichironomura (approved):

PR #1160, opened 2026-07-27
feat: add client-side authentication via configurable server headers

### LINKED ISSUE #1153: feat: client-side authentication for server access via generic header injection
## Problem

Access to the capsula server (`capsula push`, `capsula vaults`) has no authentication support. A common way to protect a self-hosted capsula-server is to put it behind an authenticating reverse proxy (Cloudflare Access, oauth2-proxy, Tailscale, etc.) — that works for browsers, but the CLI has no way to attach credentials to its requests, so the API becomes unreachable through such a proxy.

This issue proposes a generic header-injection mechanism: the CLI attaches configured HTTP headers to every request it sends to the server. Vendor integrations become documented configuration examples, with no vendor-specific code in capsula.

### Config schema

Extend the `server` field in `capsula.toml` to accept a table (the existing `server = "URL"` string form remains valid via an untagged enum):

```toml
[server]
url = "https://capsula.example.com"

[server.headers]
Authorization = { env = "CAPSULA_TOKEN", prefix = "Bearer " }
```

Each entry becomes one HTTP header line on every request the CLI sends. With `CAPSULA_TOKEN=abc123` in the environment, the example above results in:

```http
POST /api/v1/upload HTTP/1.1
Host: capsula.example.com
Authorization: Bearer abc123
```

Each header value is one of three sources:

| Form | Meaning |
|---|---|
| `"literal string"` | Used as-is |
| `{ env = "VAR", prefix = "Bearer " }` | Read from environment variable; optional `prefix` prepended |
| `{ command = "..." }` | Run the command, use trimmed stdout as the value |

This single mechanism covers common setups:

```toml
# Plain bearer token against any authenticating reverse proxy
Authorization = { env = "CAPSULA_TOKEN", prefix = "Bearer " }

# Cloudflare Access, interactive user (JWT from cloudflared login session)
cf-access-token = { command = "cloudflared access token --app=https://capsula.example.com" }

# Cloudflare Access service token (CI)
CF-Access-Client-Id = { env = "CF_ACCESS_CLIENT_ID" }
CF-Access-Client-Secret = { env = "CF_ACCESS_CLIENT_SECRET" }
```

### Decisions

- **Location**: project `capsula.toml` only — no user-level config file. `command`-based entries contain no secrets (credentials live with the external tool, e.g. `~/.cloudflared/`), so they are safe and useful to commit and share with the team. Static secrets are referenced via `env`, never written in config (consistent with the existing `dotenv` support). A `gh`-style per-host user config can be added later if a real need appears.
- **Trust model**: `command` executes arbitrary shell commands from `capsula.toml`. This follows the project's existing trust model — `capsula run` already executes arbitrary commands via the `capture-command` hook — but it newly extends that to `push`/`vaults`. Document: *config `command` entries execute arbitrary commands; only use with a trusted `capsula.toml`.*
- **Expiry UX**: no automatic re-login. If a `command` fails (e.g. expired login session of the external tool), surface the command's stderr verbatim; on HTTP 401/403, hint that credentials may be missing or expired.
- **Resolution order**: unchanged for the URL (`--server` flag > `CAPSULA_SERVER_URL` > config). Headers come from config only in this iteration.

### Implementation sketch

- `capsula-config`: `server: Option<String>` → untagged enum `Option<ServerConfig>` (string | table with `url` + `headers`)
- `capsula-client`: `CapsulaClient::new` gains resolved headers; values resolved once per client construction (command/env evaluated at that point) and attached to every request
- `capsula-cli` / `capsula-orchestration`: thread the header config through `resolve_server_url` call sites
- Docs: server access page with the configuration examples above + trust-model note

shunsuke-shimomura:
## Summary

Closes #1153

Adds client-side authentication for server access (`capsula push`, `capsula vaults`): the CLI attaches configured HTTP headers to every request it sends to the server, making the API reachable through authenticating reverse proxies (Cloudflare Access, oauth2-proxy, etc.) with no vendor-specific code in capsula.

```toml
[server]
url = "https://capsula.example.com"

[server.headers]
Authorization = { env = "CAPSULA_TOKEN", prefix = "Bearer " }
cf-access-token = { command = "cloudflared access token --app=https://capsula.example.com" }
```

- **capsula-config**: `server` accepts either the existing plain URL string (backward compatible) or a table with `url` and `headers`. Header values come from a literal string, `{ env = "VAR", prefix = "..." }`, or `{ command = "..." }`; sources mixing `env` and `command` are rejected at parse time.
- **capsula-orchestration**: new `resolve_server_headers` resolves sources at call time. Commands are split with shell-like word splitting and executed directly (no shell); failures surface the command's stderr, missing env vars name the variable.
- **capsula-client**: new `CapsulaClient::with_headers` constructor (header values marked sensitive to keep them out of debug logs); 401/403 responses get a hint that credentials may be missing or expired.
- **capsula-cli**: shared `build_server_client` helper replaces the duplicated URL-resolution/client-construction in `push` and `vaults list`. URL resolution priority is unchanged (`--server` > `CAPSULA_SERVER_URL` > config).
- **docs**: `server` / `server.headers` reference in `configuration.md` with common setups and a trust-model note (`command` entries execute arbitrary commands — same caution as the existing `capture-command` hook).

Verified with `cargo test --workspace` (159 passed, 0 failed; 15 new tests), the individual lint steps from `just lint` (clippy all-features / no-default-features, fmt, doc, check — 0 warnings), and an end-to-end smoke test against a local HTTP server confirming `Authorization: Bearer ...` (env source) and a command-sourced header arrive on the wire.

## AI assistance

- AI used: yes — implemented by Claude Code based on the design settled in #1153; a human directed the design discussion and reviewed each decision
- Human review of AI-assisted code: complete

## Breaking changes

- [x] No breaking changes
- [ ] Breaking changes; the `breaking change` label is applied

Note: the `capsula.toml` format and CLI behavior are fully backward compatible. The Rust API of `capsula-config` changes shape (`CapsulaConfig.server` is now `Option<ServerConfig>` instead of `Option<String>`) for library consumers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,6 +542,7 @@ dependencies = [
  "anyhow",
  "assert_cmd",
  "capsula-client",
+ "capsula-config",
  "capsula-core",
  "capsula-orchestration",
  "capsula-registry",
@@ -692,6 +693,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror 2.0.19",
+ "tokio",
  "wiremock",
 ]
 
@@ -753,6 +755,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "shlex",
  "tracing",
  "ulid",
  "walkdir",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -756,9 +756,12 @@ dependencies = [
  "serde",
  "serde_json",
  "shlex",
+ "tempfile",
+ "tokio",
  "tracing",
  "ulid",
  "walkdir",
+ "wiremock",
 ]
 
 [[package]]
--- a/crates/capsula-cli/Cargo.toml
+++ b/crates/capsula-cli/Cargo.toml
@@ -17,6 +17,7 @@ path = "src/main.rs"
 [dependencies]
 anyhow = { workspace = true }
 capsula-client = { workspace = true }
+capsula-config = { workspace = true }
 capsula-tui = { workspace = true }
 capsula-core = { workspace = true }
 capsula-orchestration = { workspace = true }
--- a/crates/capsula-cli/src/main.rs
+++ b/crates/capsula-cli/src/main.rs
@@ -8,7 +8,7 @@ use anyhow::{Context, Result};
 use capsula_core::hook::{PostRun, PreRun};
 use capsula_core::run::PreparedRun;
 use capsula_orchestration::push::push_single_run;
-use capsula_orchestration::resolve::resolve_server_url;
+use capsula_orchestration::resolve::{resolve_server_headers, resolve_server_url};
 use capsula_orchestration::run::{create_and_setup_run, run_post_hooks, run_pre_hooks};
 use capsula_orchestration::setup::LoadedConfig;
 use capsula_orchestration::vault::{find_run_dir, find_run_dir_by_name, list_runs};
@@ -394,14 +394,7 @@ path = \".\"
             all,
             server,
         } => {
-            let server_url =
-                resolve_server_url(server, config.server.as_deref()).ok_or_else(|| {
-                    anyhow::anyhow!(
-                        "Server URL not specified. Use --server <URL>, set CAPSULA_SERVER_URL environment variable, or add 'server = \"URL\"' to capsula.toml"
-                    )
-                })?;
-
-            let client = capsula_client::CapsulaClient::new(&server_url);
+            let (server_url, client) = build_server_client(server, config.server.as_ref())?;
             let vault_name = &config.vault.name;
 
             match client.vault_exists(vault_name) {
@@ -488,16 +481,10 @@ path = \".\"
         Commands::Tui => unreachable!("Handled above"),
         Commands::Vaults { command } => match command {
             VaultsCommands::List { server } => {
-                let server_url =
-                    resolve_server_url(server, config.server.as_deref()).ok_or_else(|| {
-                        anyhow::anyhow!(
-                            "Server URL not specified. Use --server <URL>, set CAPSULA_SERVER_URL environment variable, or add 'server = \"URL\"' to capsula.toml"
-                        )
-                    })?;
+                let (server_url, client) = build_server_client(server, config.server.as_ref())?;
 
                 info!("Fetching vaults from server {}", server_url);
 
-                let client = capsula_client::CapsulaClient::new(&server_url);
                 let vaults = client.list_vaults().context("Failed to list vaults")?;
 
                 if vaults.is_empty() {
@@ -516,6 +503,30 @@ path = \".\"
         },
     }
     Ok(())
+}
+
+/// Resolve the server URL and configured headers, and build an HTTP client.
+fn build_server_client(
+    server_override: Option<String>,
+    config_server: Option<&capsula_config::ServerConfig>,
+) -> Result<(String, capsula_client::CapsulaClient)> {
+    let server_url = resolve_server_url(server_override, config_server.map(|s| s.url.as_str()))
+        .ok_or_else(|| {
+            anyhow::anyhow!(
+                "Server URL not specified. Use --server <URL>, set CAPSULA_SERVER_URL environment variable, or add 'server = \"URL\"' to capsula.toml"
+            )
+        })?;
+
+    let headers = config_server
+        .map(|s| resolve_server_headers(&s.headers))
+        .transpose()
+        .context("Failed to resolve server headers")?
+        .unwrap_or_default();
+
+    let client = capsula_client::CapsulaClient::with_headers(&server_url, &headers)
+        .context("Failed to build server client")?;
+
+    Ok((server_url, client))
 }
 
 fn read_json_if_exists(path: &std::path::Path) -> Result<Option<serde_json::Value>> {
--- a/crates/capsula-cli/src/main.rs
+++ b/crates/capsula-cli/src/main.rs
@@ -8,7 +8,7 @@ use anyhow::{Context, Result};
 use capsula_core::hook::{PostRun, PreRun};
 use capsula_core::run::PreparedRun;
 use capsula_orchestration::push::push_single_run;
-use capsula_orchestration::resolve::{resolve_server_headers, resolve_server_url};
+use capsula_orchestration::resolve::{is_same_origin, resolve_server_headers, resolve_server_url};
 use capsula_orchestration::run::{create_and_setup_run, run_post_hooks, run_pre_hooks};
 use capsula_orchestration::setup::LoadedConfig;
 use capsula_orchestration::vault::{find_run_dir, find_run_dir_by_name, list_runs};
@@ -83,6 +83,11 @@ enum Commands {
         #[arg(long)]
         json: bool,
     },
+    #[command(group(
+        clap::ArgGroup::new("push_target")
+            .required(true)
+            .args(["run_id", "all"])
+    ))]
     Push {
         /// Run ID or name to push (e.g., 01HQXYZ... or chubby-back)
         run_id: Option<String>,
@@ -394,7 +399,8 @@ path = \".\"
             all,
             server,
         } => {
-            let (server_url, client) = build_server_client(server, config.server.as_ref())?;
+            let (server_url, client) =
+                build_server_client(server, config.server.as_ref(), &project_root)?;
             let vault_name = &config.vault.name;
 
             match client.vault_exists(vault_name) {
@@ -445,7 +451,7 @@ path = \".\"
                             continue;
                         }
 
-                        match push_single_run(run_dir, vault_name, &server_url, &client) {
+                        match push_single_run(run_dir, vault_name, &client) {
                             Ok(()) => success_count += 1,
                             Err(e) => {
                                 if e.to_string().contains("already exists") {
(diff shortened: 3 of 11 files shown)


shunichironomura:
Thank you for the quick PR.

Below is the review from GPT-5.6 Sol (thinking: max). In general I think all findings are valid and worth addressing.

---

> [!WARNING]
> This review is generated by AI and needs human review. 

# `feat/server-auth-headers` code review

Branch: `feat/server-auth-headers`  
PR: <https://github.com/ut-issl/capsula/pull/1160>  
Reviewed commit: `e212e7ec6a17edf58b84ca288dc8a1463ad5ae83`  
Review base: `bf4b0c40d6d5596f695e70224946495a90d2b428` (`v0.14.0`)

## Summary

The configuration model is backward-compatible at the TOML level, the header-source enum is appropriately typed, credential values are not logged during resolution, and the CLI client-construction duplication was reduced. The command source also avoids invoking a shell, and the documentation calls out the new trusted-config boundary.

The branch is not ready to merge, however. The main `capsula push` metadata request still constructs a separate unauthenticated HTTP client, so the feature does not actually authenticate the complete push flow. In addition, reqwest's default redirect behavior forwards arbitrary configured headers such as Cloudflare service-token headers across origins. Marking a `HeaderValue` as sensitive does not prevent that forwarding.

## Findings

### 1. Blocker — `capsula push` sends the create-run request without configured headers

**Locations:**

- `crates/capsula-orchestration/src/push.rs:66-69`
- `crates/capsula-orchestration/src/push.rs:93-98`
- `crates/capsula-cli/src/main.rs:397`

`build_server_client` constructs the header-bearing `CapsulaClient`, and `push_single_run` uses it for `/api/v1/upload`. The preceding metadata request to `POST /api/v1/runs`, however, bypasses that client:

```rust
let http_client = reqwest::blocking::Client::new();
let response = http_client.post(&url).json(&create_run_payload).send()?;
```

Consequences behind an authenticating reverse proxy:

- a direct 401/403 makes every push fail before upload, without the new authentication hint;
- a redirect to a login page can end in HTTP 200, which this code treats as a successful run creation without inspecting the response body;
- if the run has no non-`_capsula` files or hook output, upload is skipped and Capsula can report `Pushed` even though no run was created.

The current tests verify headers on `list_vaults`, but no test exercises the complete push request sequence, which allowed this bypass to remain unnoticed.

**Recommendation:** Move run creation into `CapsulaClient` so it uses the same underlying reqwest client and base URL as vault lookup and upload. Prefer a typed `Created | AlreadyExists` outcome for the expected 201/409 statuses rather than exposing another raw client. Remove the separate `server_url` request path from `push_single_run`, and add an end-to-end test that requires the configured header on `/api/v1/vaults/{name}`, `/api/v1/runs`, and `/api/v1/upload`.

### 2. High — Custom credentials are forwarded across cross-origin redirects

**Locations:**

- `crates/capsula-client/src/lib.rs:44-66`
- `crates/capsula-client/src/lib.rs:85-86,101-102,123,161`

`CapsulaClient::with_headers` installs credentials as reqwest default headers while retaining reqwest's default redirect policy. Reqwest 0.13.4 follows up to ten redirects and, on a cross-origin redirect, removes only known names such as `Authorization`, `Cookie`, and `Proxy-Authorization`. It does not remove generic credentials explicitly supported by this branch, including `CF-Access-Client-Id`, `CF-Access-Client-Secret`, and `cf-access-token`.

`HeaderValue::set_sensitive(true)` masks debug output and can affect caching/compression behavior; it is not an origin-binding or redirect policy. The locked reqwest behavior is visible in its [redirect implementation](https://docs.rs/reqwest/0.13.4/src/reqwest/redirect.rs.html#239-250).

A local two-server probe confirmed the impact: server A returned a 302 to server B, and B received a configured `CF-Access-Client-Secret: top-secret` header; `CapsulaClient::list_vaults` completed successfully against B.

This can disclose a service token to an identity provider, an accidentally configured redirect target, or an attacker-controlled target selected by a compromised server response.

**Recommendation:** Do not follow redirects for authenticated API requests, or permit only redirects whose scheme, host, and effective port match the configured server origin. Add cross-origin and same-origin redirect regression tests for a custom credential header. If a cross-origin login redirect is stopped, surface it as an authentication/configuration error rather than following it to HTML and producing a JSON decode error.

### 3. Medium — Configured credentials are also sent to unrelated URL overrides

**Locations:**

- `crates/capsula-cli/src/main.rs:509-526`
- `crates/capsula-orchestration/src/resolve.rs:50-73`
- `docs/configuration.md:55-72`

The URL is selected with `--server > CAPSULA_SERVER_URL > [server].url`, but headers are always resolved from the config table and attached to the selected URL. Therefore this configuration:

```toml
[server]
url = "https://prod.example"

[server.headers]
CF-Access-Client-Secret = { env = "PROD_SECRET" }
```

will send `PROD_SECRET` to any origin supplied through `--server` or `CAPSULA_SERVER_URL`. A command source can similarly generate a token for the configured production application and then send it to the override target. The issue deliberately preserves URL precedence, but neither the implementation nor documentation establishes this credential-scoping consequence.

**Recommendation:** Bind configured headers to the configured server origin. Reject a cross-origin override when headers are present, or omit the headers unless the selected URL has the same origin. If sending config credentials to arbitrary overrides is intentionally retained, require explicit opt-in and document the disclosure risk. Add tests for CLI and environment overrides.

### 4. Medium — Typos in the security-sensitive `[server]` table are silently ignored

**Location:** `crates/capsula-config/src/lib.rs:84-92`

The detailed deserialization helper does not use `#[serde(deny_unknown_fields)]`. A typo such as `heders = {...}` or `header = {...}` is accepted as long as `url` is present; Capsula then silently sends no credentials. This is inconsistent with the strict `EnvHeaderSource` and `CommandHeaderSource` variants and with the project's normal hook-configuration behavior.

**Recommendation:** Deserialize the detailed form through a dedicated `#[serde(deny_unknown_fields)]` struct and add a regression test asserting that a misspelled server field fails configuration loading.

### 5. Low — Header commands run in the caller's working directory, not the config project root

**Locations:**

- `crates/capsula-orchestration/src/resolve.rs:116-126`
- `crates/capsula-cli/src/main.rs:520-523`

`run_header_command` invokes `Command::new(...).args(...).output()` without setting `current_dir`, and `resolve_server_headers` receives no project-root path. A shared config using `command = "sh scripts/get-token.sh"` therefore behaves differently depending on where Capsula is launched; with `--config /project/capsula.toml`, it can run a same-named relative script from an unrelated directory. The existing `capture-command` hook resolves its default working directory to the project root, so this new config-driven command behaves inconsistently.

**Recommendation:** Pass `project_root` through header resolution and set it as the helper's working directory, or add an explicit typed `cwd` option and clearly document the default. Cover invocation from a directory outside the config root.

### 6. Low — An invalid `capsula push` invocation executes credential helpers before argument validation

**Location:** `crates/capsula-cli/src/main.rs:397-415`

When neither a run ID nor `--all` is supplied, Capsula builds the client, executes all command-backed header sources, and performs `vault_exists` before reaching the runtime error that requires one of those arguments. This unnecessarily adds arbitrary-command and network side effects to an invalid invocation.

**Recommendation:** Express the requirement as a Clap argument group if possible, or validate `(all, run_id)` before constructing the server client and contacting the server.

## Implementation plan

1. Add authenticated run creation to `CapsulaClient`; route every Capsula API request through that client and add a full push-flow regression test.
2. Establish an origin policy for credentials: disable redirects or allow only same-origin redirects, then test custom headers across redirect boundaries.
3. Bind headers to the configured origin when CLI/environment URL overrides are used; document the selected policy.
4. Make detailed server configuration strict and run command-backed sources from a defined project-root working directory.
5. Validate push target arguments before resolving credentials or performing network I/O.
6. Update from current `origin/main` (the reviewed branch is seven commits behind), rerun the complete validation suite, and repeat the end-to-end proxy smoke test.

## Validation performed

Passed locally at `e212e7ec`:

- `cargo test --workspace` — 159 passed, 0 failed; 2 doctests ignored
- `just lint` — both Clippy configurations, formatting, docs, and workspace check passed
- `git diff --check e212e7ec^ e212e7ec`

Additional review probes/audits:

- audited all non-server `/api/v1/*` request construction and found the unauthenticated `/api/v1/runs` client in `push.rs`;
- inspected the locked reqwest 0.13.4 redirect implementation;
- reproduced forwarding of a sensitive custom header across a cross-origin 302 with two local wiremock servers.

## Positive notes

- Existing `server = "URL"` files remain valid.
- `HeaderValueSource` models literal, environment, and command sources explicitly, and mixed source tables are rejected.
- Header values are omitted from resolver logs and marked sensitive in the HTTP client.
- Commands are executed directly rather than through a shell, with useful missing-variable and command-failure context.
- The shared CLI constructor removes duplicated URL/client setup for `push` and `vaults list`.
- The documentation appropriately warns that command-backed sources extend the trusted-config boundary to `push` and `vaults`.

shunsuke-shimomura:
Thank you for the thorough review — all six findings were valid. Addressed in 012e6a3 (also rebased onto current `main`):

1. **Blocker — unauthenticated `/api/v1/runs` request**: run creation moved into `CapsulaClient::create_run` with a typed `Created | AlreadyExists` outcome; `push_single_run` no longer takes a `server_url` and the raw `reqwest` client is gone. New end-to-end test drives the full push flow against a mock that *requires* the configured header on both `/api/v1/runs` and `/api/v1/upload` (plus a client-level header test for `/api/v1/vaults/{name}`).
2. **High — credential forwarding across redirects**: the client now uses `redirect::Policy::none()` (both constructors). A 3xx response surfaces as an error with a hint that it may be a login redirect. Regression test: a 302 to a second wiremock server yields an error and the target receives zero requests.
3. **Medium — credentials sent to URL overrides**: `[server.headers]` are now bound to the configured origin. A `--server`/`CAPSULA_SERVER_URL` override with a different origin (scheme/host/effective port, compared via `Url::origin()`) is rejected with an explicit error. Documented in `configuration.md`.
4. **Medium — silent typos in `[server]`**: the detailed form now deserializes through a `#[serde(deny_unknown_fields)]` struct; a misspelled field fails configuration loading (regression test included).
5. **Low — command working directory**: command-backed sources now run with the project root as `current_dir`, matching the `capture-command` hook (test included).
6. **Low — side effects on invalid `capsula push`**: the run-ID/`--all` requirement is now a required clap `ArgGroup`, so an invalid invocation fails at argument parsing before any credential helper or network I/O runs (verified with a `touch`-based smoke test).

Validation: `cargo test --workspace` 169 passed / 0 failed; both clippy configurations, `fmt --check`, `doc`, `check` clean; end-to-end smoke tests for origin rejection and the clap group rerun locally.

